PR #10, opened 2014-10-11
Complete rewrite

martinthomson:
@fluffy @keithgrif

Don't merge this pull request just yet, it's not ready, but this is an attempt to approach this differently and examine how this might better fit into our use cases.  I need to flesh out a lot of this.

I'm working on getting the authorization pieces of this, I'm thinking that the terminology and taxonomy that I've developed here should help that a lot.

This depends on #8.

### DIFF
--- a/screenshare.css
+++ b/screenshare.css
@@ -1,7 +1,7 @@
 
 @media screen {
   html { background: #eeeeee; }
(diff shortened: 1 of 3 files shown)


martinthomson:
OK, I'm happy with this now.  I'd like to get feedback from @fluffy and @keithgrif before I go any further.

PR #23, opened 2015-02-07
Change the name to getDisplayMedia

### LINKED ISSUE #15: getOutputMedia() is a confusing name
Perhaps:
getDisplayMedia()?

martinthomson:
As suggested in #15.

Closes #15.

### DIFF
--- a/index.html
+++ b/index.html
@@ -21,7 +21,7 @@
 <body>
   <section id="abstract">
     <p>This document defines how a user's display, or parts thereof, can be used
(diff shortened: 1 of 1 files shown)


martinthomson:
@dontcallmedom, the whole travis thing is clearly flaky.  Is there something we can do about that?

dontcallmedom:
@martinthomson I suggest disabling the linkchecking rule in travis.yml while I figure why it fails on IETF URLs (which I won't be able to get to before a week or two)

PR #34, opened 2015-11-24
Transparency hazards

martinthomson:


index.html
```diff
@@ -167,78 +165,88 @@
...
+      constraints.</p>
 
       <section>
-        <h2><code>OutputCaptureSurfaceType</code></h2>
+        <h2>New Constraints for Captured Display Surfaces</h2>
+        <p>Two constraints are defined that allow an application to observe
+        properties of the selected <a>display surface</a>.  Since the source of
+        media cannot be changed after a <code>MediaStreamTrack</code> has been
+        returned, these constraints cannot be changed by an application.</p>
+
```
alvestrand, line 193: This intro paragraph is not consistent with the description of the "logicalSurface" constraint below, since that is a preference, not just an observable property.

martinthomson, line 193: Good point.  I think that the cleanest model here is to view the logicalSurface "constraint" as a selecting constraint, in the same way that facingMode is used to select.  Once set, it can't be changed.  I've opened #36 to track this one, since I'm a little behind on other things and getting this right will be hard.

PR #44, opened 2016-10-05
Disable Local Echo

chenqiang1986:


dontcallmedom:
@chenqiang1986 I understand you work for Google; would you consider [requesting to be added to the Working Group](https://www.w3.org/2004/01/pp-impl/47318/join)?

chenqiang1986:
Any further action? How can I link W3C account with github account?

plehegar:
@chenqiang1986 you're all set. Your account has been linked (it's done manually so far unfortunately).

chenqiang1986:
Whom should I notify for the review?

martinthomson:
Discussion of this spec happens primarily on public-media-capture@w3.org.  If you want to start a thread there, that would be a good way to start.

chenqiang1986:
I started a thread in public-media-capture@w3.org months ago, but no reply at all.
Anything else I can do?

martinthomson:
Maybe you could convince the group that this is worth doing.  I don't think that this is a good idea, but I'm always open to arguments about why something might be useful.

alvestrand:
The next question is "what's your use case". The mail to the TF list was better at explaining it than this issue itself. Do you want to repeat it here?

index.html
```diff
@@ -201,6 +204,15 @@ <h2 id="constraints">Constraining Display Surface Selection</h2>
         the <a>visible display surface</a>.</p>
 
         <dl class="idl" title="partial dictionary MediaTrackConstraintSet">
+          <dt>ConstrainBoolean disableLocalEcho</dt>
+
+          <dd>
+            <p>A value of <code>true</code> indicates that we mute the local
+            playback for the captured content. For example, when capturing a
+            browser tab, the tab will be muted then; when capturing the full
+            screen with audio, then the system audio will be muted then.</p>
```
alvestrand: Grammar nit: I think you can delete all occurences of "then" in this sentence and make it more readable.

index.html
```diff
@@ -186,11 +186,14 @@ <h2 id="constraints">Constraining Display Surface Selection</h2>
...
       <section>
         <h2>New Constraints for Captured Display Surfaces</h2>
-        <p>Two constraints are defined that allow an application to observe
+        <p>Three constraints are defined that allow an application to observe
         properties of the selected <a>display surface</a>.  Since the source of
         media cannot be changed after a <code>MediaStreamTrack</code> has been
         returned, these constraints cannot be changed by an application.</p>
 
+        <p>The <code><a href="#widl-MediaTrackConstraintSet-disableLocalEcho">disableLocalEcho</a></code>
+        constraint allows an application to suppress local playback of the captured audio.</p>
```
alvestrand: A name of "muteAudioForSource" might be easier to understand. Having booleans with "disable" in the name is usually bad practice.

index.html
```diff
@@ -186,11 +186,14 @@ <h2 id="constraints">Constraining Display Surface Selection</h2>
 
       <section>
         <h2>New Constraints for Captured Display Surfaces</h2>
-        <p>Two constraints are defined that allow an application to observe
+        <p>Three constraints are defined that allow an application to observe
         properties of the selected <a>display surface</a>.  Since the source of
         media cannot be changed after a <code>MediaStreamTrack</code> has been
         returned, these constraints cannot be changed by an application.</p>
```
alvestrand, line 193: This description seems not accurate any more, since the new constraint modifies a property of the captured surface. Can you rephrase?

alvestrand (commented):
This doesn't seem an unreasonable thing to ask for, but it's not completely done yet.
You will also want to mention what you expect the normal behavior (no constraint) to be - I'd expect it to be that the captured surface's associated sound continues to play locally - that is, "disableLocalAudio=false".

chenqiang1986:
We are actually driven by the use cases from users, as we found without such freedom, either default setting (mute or unmute), some users will be unhappy.

1. For peer to peer connection, for example hangouts session or any other video chat, when sharing audio, the user would in general expect that both the sender and receiver can hear the sound.

2. For Cast, or any other app that is designed to project a laptop to the big screen in the conference room, we would hope that only the receiver (the speaker associated with the big screen) would play out the audio.

chenqiang1986:
Thanks for review. Rephrased.

index.html
```diff
@@ -217,6 +221,15 @@ <h2>New Constraints for Captured Display Surfaces</h2>
...
           </dd>
+
+          <dt>ConstrainBoolean muteAudioForSource</dt>
+
+          <dd>
+            <p>A value of <code>true</code> indicates that we mute the local
+            playback for the captured content. For example, when capturing a
+            browser tab, the tab will be muted; when capturing the full screen
+            with audio, then the system audio will be muted.</p>
+          </dd>
```
henbos, line 232: For tab capture, does "The tab will be muted" mean everything is silence, or does this mean we get all system audio except for that tab?

When capturing the full screen, is muting the system audio the same as not requesting audio?

What do you think about this proposal?
https://docs.google.com/document/d/1pgN8Pe2zMaD0tytJVsKmnBT-7gM247YlzzWazm7OwSk/edit?usp=sharing
It overlaps with some of this.

henbos:
I filed #79

alvestrand:
Closing as OBE.

PR #46, opened 2016-10-10
Clarifying constraints and their application

martinthomson:
This is a fairly large change, intended to better address a bunch of open issues around the use of constraints.

The algorithm change is actually simple, despite what shows in the diff here: the constraints are passed to the getUserMedia() algorithm but `audio` and `video` are replaced with a simple Boolean.  If a MediaStreamTrackConstraints value was replaced, then applyConstraints is called on the resulting track.  The only downside here being that you potentially get an overconstrained (and therefore empty) track, with no event (since that happens before you have the track).

Also here, I made the argument optional and default to `{video:true}`, since that's the most sensible thing to want to do.

@jan-ivar, @stefhak, do you think that MediaStreamTrack should have an overconstrained attribute?

Attention: @ShijunS who asked for this.


stefhak:
@martinthomson you mean an overconstrained attribute in addition to the event? To be specced in the mediacapture-main doc?

jan-ivar:
@martinthomson Do we expect users to do this:

``` js
let stream = await getDisplayMedia({video: {width: {max: 320}, height: {max: 200}}});
if (stream.getVideoTracks()[0].overconstrained) {
    throw new OverconstrainedError("No dice");
}
```

?

That seems bad, as it's easy to forget to check for `overconstrained`. This seems preferable:

``` js
let stream = await getDisplayMedia({video: true});
await stream.getVideoTracks()[0].applyConstraints({width: {max: 320}, height: {max: 200}});
```

If we don't support selection through constraints, the API should be upfront about that IMHO. I'd prefer if the top call to `getDisplayMedia` rejected with `TypeError`.

index.html
```diff
@@ -164,210 +164,257 @@
         "#constraints"></a>). Two new constraints allow an application to observe what type of
         information was acquired.
       </p>
+      <div class="note">
+        <p>
+          Though all constraints are supported for display device capture, not all constraints will
+          have an effect. Usually, this is because an output or display device cannot alter its
+          behavior in a way that the constraint implies, just as a video camera cannot alter
+          <a href="https://w3c.github.io/mediacapture-main/#def-constraint-volume">volume</a>.
```
jan-ivar: I think we need to write to spec implementers, not users here. It's this specification's job to nail down which constraints apply to screen sharing IMHO.

I'm not a fan of the "if it's in MediaTrackConstraintSet, it may do something useful with screensharing, or it may not". For this to be a compliance document, we should avoid implying functionality.

alvestrand:
Not reading the diff, but noting that getUserMedia() doesn't reply with an overconstrained event; it returns a promise, and if the result is overconstrained, the promise is rejected with an argument that is an OverconstrainedError.

I don't see a reason to depart from that method here.

So Jan-Ivar's example:

let stream = await getDisplayMedia({video: {width: {max: 320}, height: {max: 200}}});
if (stream.getVideoTracks()[0].overconstrained) {
    throw new OverconstrainedError("No dice");
}

would be simply

async function MyFunction() {
  let stream = await getDisplayMedia({video: {width: {max: 320}, height: {max: 200}}});
}

since a rejected promise in await returns as a rejection of the promise implicitly retured by an async function.

(If I've understood everythign correctly)

jan-ivar:
@alvestrand Works, but the difference is `getUserMedia` fails without bugging the user, whereas `getDisplayMedia` would always bug the user, and then fail, which seems undesirable.

index.html
```diff
@@ -164,210 +164,257 @@
...
+        <p>
+          For instance, it is unlikely that setting the following constraints will have any useful
+          effect (other than an <a href=
+          "https://w3c.github.io/mediacapture-main/#event-mediastreamtrack-overconstrained">overconstrained</a>
+          error): <a href=
+          "https://w3c.github.io/mediacapture-main/#def-constraint-aspectratio">aspectRatio</a>,
+          <a href=
+          "https://w3c.github.io/mediacapture-main/#def-constraint-facingmode">facingMode</a>, and
+          <a href=
+          "https://w3c.github.io/mediacapture-main/#def-constraint-echocancellation">echoCancellation</a>.
```
jan-ivar, line 183: I think handling of irrelevant constraints should be dealt with in the media capture spec instead. Issue https://github.com/w3c/mediacapture-main/issues/409 filed.

index.html
```diff
@@ -164,210 +164,257 @@
...
-          </section>
-        </div>
+        <dl data-link-for="NavigatorUserMedia" data-dfn-for="NavigatorUserMedia" class="methods">
+          <dt>
+            <code><dfn data-dfn-for="getDisplayMedia">getDisplayMedia</dfn></code>
+          </dt>
+          <dd>
+            <p>
+              This method operates identically to <code>getUserMedia</code>, except that it
+              acquires media from output devices.
```
jan-ivar: Not new to this patch, but I don't see how a screen is an "output device" and a camera is not in our context.

Media capture already uses "output" to talk about sinks (e.g. speakers), but here we're using it to talk about sources (e.g. audio capture from a screen or window).

I'd favor a script-centric model ("input" means source, "output" means sink).

```
inputs (cameras, mics, screens, windows) -> JS -> outputs (elements, recorder, speakers, headphones)
```

index.html
```diff
@@ -378,12 +425,16 @@ <h2 id="constraints">
           system.
         </p>
         <p>
-          Display capture sources MUST NOT be selected with the <code>deviceId</code> constraint,
-          since this would allow applications to influence selection. A display capture source is
-          represented in the <code>MediaStreamTrack</code> API as having a <code>deviceId</code>
-          parameter that is randomized each time a <code>MediaStreamTrack</code> is connected. The
-          only other constraint on this value is that it cannot duplicate any existing values for
-          <code>deviceId</code>.
+          Display capture sources cannot be selected with the <code>deviceId</code> constraint,
```
jan-ivar: This replaces normative MUST NOT language (speaking to implementer) with passive description of behavior (speaking to user?) Is this in reference to a normative guarantee given elsewhere? If not, we should restore the MUST NOT, otherwise the statement doesn't seem true.

index.html
```diff
@@ -418,9 +469,9 @@ <h2 id="constraints">
       </p>
       <p>
         Display capture that includes browser windows, particularly those that are under any form
-        of control by the application, risks violation of these basic security protections. Though
-        this risk is not entirely contained to browser windows, since control channels between
-        browser applications and other applications, depending on the operating system. The key
+        of control by the application, risks violation of these basic security protections. This
+        risk is not entirely contained to browser windows, since control channels between browser
+        applications and other applications, depending on the operating system. The key
```
jan-ivar, line 462: Is "channels" a verb here?

index.html
```diff
@@ -164,210 +164,257 @@
...
+        </p>
+        <p>
+          If either of the <code><a href=
+          "https://w3c.github.io/mediacapture-main/#dom-mediastreamconstraints-video">video</a></code>
+          and <code><a href=
+          "https://w3c.github.io/mediacapture-main/#dom-mediastreamconstraints-audio">audio</a></code>
+          attributes of the <code>constraints</code> argument have a value that is an instance of
+          <code><a href=
+          "https://w3c.github.io/mediacapture-main/#idl-def-mediatrackconstraints">MediaTrackConstraints</a></code>,
+          that value is replaced with a simple Boolean <code>true</code>. All other attributes are
```
jan-ivar: s/a simple Boolean true/true/ imho.

jan-ivar (commented):
Great to see progress on this!

I have some concerns, mostly to do with functionality implied by the mediacapture spec and the constrainable pattern.

ShijunS:
The [proposal in Sapporo](https://www.w3.org/2011/04/webrtc/wiki/images/2/24/Capture_items.pdf) (slide 3) was to remove the new constraints and let user choose the capture surface.  I remember we supported the proposal.  Can anyone else confirm the WG/TF consensus?

jan-ivar:
@alvestrand To clarify what I mean: The following - a good pattern in gUM - is a bad pattern here:

``` js
let stream;
try {
  stream = await getDisplayMedia({video: {width: {max: 320}, height: {max: 200}}});
} catch (e) {
  if (e.name == "OverconstrainedError") {
    stream = await getDisplayMedia({video: fallback_constraints});
  }
}
```

because unlike gUM, it prompts the user twice. Contrast with the other way which doesn't:

``` js
let track = (await getDisplayMedia({video: true })).getVideoTracks()[0];
try {
  await track.applyConstraints({video: {width: {max: 320}, height: {max: 200}}});
} catch (e) {
  if (e.name == "OverconstrainedError") {
    await track.applyConstraints(fallback_constraints);
  }
}
```

jan-ivar:
@ShijunS I can't speak to the consensus, but I would gladly support that. Constraints is a terrible API surface for a rescaler (if we want a rescaler).

alvestrand:
Can't we fix the bug that GetDisplayMedia can fail after prompting instead?

wrt applicability of constraints: At least the way I've coded it, if you apply an irrelevant constraint (volume to a video track) as (exact, max, min), it will cause the call to fail. If you apply an optional constraint, the UA is free to ignore it, so it makes sense to ignore it (this gives future-proofing for constraints where the app wants to use it if it's present, but doesn't care if it's not present).

jan-ivar:
@alvestrand You want `getDisplayMedia` to fail upfront unless all possible sources satisfy the given constraints?

Re: applicability of constraints, Firefox does not fail in that case, though we should discuss that in https://github.com/w3c/mediacapture-main/issues/409.

index.html
martinthomson: I'm merely being consistent with the existing API, namely that if you specify "volume" on video, you get what you deserve (i.e., overconstrained, I think).

index.html
martinthomson: That works too.

index.html
martinthomson: How is this not script-centric?  In your taxonomy, camera == input, screen == output.  This is special in that it closes the loop (and makes an output device an input as well).

martinthomson:
@ShijunS 

> The proposal in Sapporo (slide 3) was to remove the new constraints and let user choose the capture surface. I remember we supported the proposal. Can anyone else confirm the WG/TF consensus?

Yes, we agreed not to let the application select the surface.  However, removing constraints entirely plays poorly with other attributes on `MediaStreamConstraints`, such as `peerIdentity`.  And I believe that it is still useful to support `{ audio: true, video: true }` as opposed to just `{ video: true }`.  Hence this proposal.

@alvestrand 

> Can't we fix the bug that GetDisplayMedia can fail after prompting instead?

Not sure what you want here.  getUserMedia filters the sources based on constraints, asks the user to pick from the remainder, and returns the result (or an error if the user cannot or does not pick).  If we change the getDisplayMedia to ask the user to pick, then fail if the result isn't OK, that leads to some of the strange behaviour @jan-ivar pointed at.

Based on this discussion, would people be OK if I revised this to throw if `audio` or `video` include a non-Boolean value?  That would be the simplest solution I think.

index.html
jan-ivar: No, "screen" is an input. It's a source, not a sink. I don't think I care that it is also something else in other contexts. YMMV.

index.html
martinthomson: It's not a replacement.  The previous paragraph establishes the requirement on implementers with a MUST NOT: you can't enumerate devices using `enumerateDevices`.  This is just explaining the consequences of that choice.

martinthomson:
Updated to reject, plus a few small improvements.  I will revise this based on the conclusion to w3c/mediacapture-main#409, or, if it's acceptable, merge it and then open another issue to track progress over there.

index.html
```diff
@@ -164,210 +164,244 @@
...
-          </section>
-        </div>
+        <dl data-link-for="NavigatorUserMedia" data-dfn-for="NavigatorUserMedia" class="methods">
+          <dt>
+            <code><dfn data-dfn-for="getDisplayMedia">getDisplayMedia</dfn></code>
+          </dt>
+          <dd>
+            <p>
+              This method operates identically to <code><a>getUserMedia</a></code>, except that it
+              acquires media from display devices and the <code>constraints</code> argument does
```
jan-ivar: comma?

index.html
```diff
@@ -164,210 +164,244 @@
...
+              not accept <code><a href=
+              "https://w3c.github.io/mediacapture-main/#idl-def-mediatrackconstraints">MediaTrackConstraints</a></code>
+              values. This prevents an application from influencing the section of sources, see
```
jan-ivar: Did you mean selection?

jan-ivar:
Lgtm.

PR #47, opened 2017-04-27
remove the outdated paragraph

munrochiang:
Since we decided to reduce the gDM constraint to a simple boolean, we don't need this paragraph


martinthomson:
Thanks!

PR #50, opened 2017-10-20
Upgrade to NodeJS 8 (respec 1.7 uses async/await)

vivienlacourba:


martinthomson:
Hard to make sense of the error here.

Issue #29, opened 2015-09-10
Fullscreen needs handling (was: Powerpoint is special)


martinthomson:
We will have to detect when a particular Powerpoint window has become full screen and shift the window share to the presentation.  We might want to extend the same sort of privilege to other applications that shift to a full screen mode.

aboba:
Not clear this is needed if the user can select the monitor that the presentation is displayed on.

mhofman:
It is needed if the user is able to select application / window, and that application / window transitions to a full screen rendering.

We are getting a ton of complaints from users that start sharing "Powerpoint" because we allow selecting windows as a source, and subsequently start presentation mode and attendees only see a black screen.

If only full screens are allowed to be selected as a source, this is not an issue, but since this API does not allow the website to constraint which source the user can select to work around the issue, the user agent should be smart enough to respect the user's choice of "sharing the powerpoint window".

ShijunS:
> the user agent should be smart enough to respect the user's choice of "sharing the powerpoint window".

Yes, it'd make sense to leave it as a UA implementation decision in order to deliver a seamless user experience.  As an example, for a specific app, e.g. PowerPoint, the UA can always chose to stream content out of a "primary" display surface of the app during runtime.  The UA can dynamically decide which display surface is the "primary" one during runtime, in case there are multiple to select from.

mhofman:
I think the idea is to specify or at least document this behavior to avoid un-optimal implementations like we currently have?

ShijunS:
Yes, a non-normative implementation suggestion would help.

dot-slash-s:
Update from Feb 21st meeting 
Write a proposal based on the following lines
"""
Applications that go full-screen might need special treatment
Modes for consideration:
    “application” DisplayCaptureSurfaceType selected
    “window” DisplayCaptureSurfaceType selected
In either case, if the UA is able to make unambiguous 1:1 determination that the invisible window displayed in full screen mode maps to the window (initially selected) from the same application, Screen Capture should be allowed.
Proposal: Write informal note to explain the above and consequences of mapping failure.
"""

henbos:
Fullscreen is a platform-specific concept, we decided at TPAC to have informative language but nothing normative. "Fullscreen" is not a special mode.

eladalon1983:
This issue has not seen any activity in a long time. It appears that, important though it is, it is not important enough to merit action. I am closing this. Feel free to reopen.

PR #57, opened 2018-04-12
Issue 29 - Powerpoint is Special

suhasHere:
Here is the first pass at providing some non-normative recommendation when capturing full screen mode applications.

### DIFF
--- a/index.html
+++ b/index.html
@@ -288,7 +288,7 @@
             <dd>
               <p>
                 A value of <code>true</code> indicates capture of a <a>logical display surface</a>;
-                a value of <code>false</code> indicates a capture capture of a <a>visible display
+                a value of <code>false</code> indicates a capture of a <a>visible display
                 surface</a>.
               </p>
             </dd>
@@ -379,6 +379,32 @@
           duplicate any existing values.
         </p>
       </section>
+      <section>
+        <h2 id="fullscreen">
+          Fullscreen Rendering considerations
+        </h2>
+          <p><i>This section is non-normative.</i></p>
+          <p>
+            For applications that support full screen rendering such as powerpoint applications,
+            the decision of the display surface that gets captured while in fullscreen mode is 
+            specific to the user agent implementation. A user agent MAY choose to switch from 
+            capture of a window to the fullscreen when the associated application goes 
+            fullscreen. For consistency reasons, an user agent must ensure that the display surface 
+            captured in fullscreen mode be unambiguously mapped to the application/window where 
+            the action was triggered in the first place. This might be achieved by ensuring that 
+            this transition is only made when the captured window has focus before the transition.
+          </p>
+          <p>
+            The choice of <code><a>DisplayCaptureSurfaceType</a></code> and 
+            <code><a data-link-for="MediaTrackConstraintSet">logicalSurface</a></code> setting 
+            may further influence what gets captured. For example, when an application/window 
+            is triggered to go fullscreen with the
+            <code><a data-link-for="MediaTrackConstraintSet">logicalSurface</a></code> value 
+            of false and the DisplayCaptureSurfaceType of "monitor", the choice of the "monitor" 
+            <a>visible display surface</a> selected for capture is solely left to the user agent, 
+            especially when there are multiple options available.
+          </p> 
+      </section>
     </section>
     <section>
       <h2>
@@ -417,6 +443,14 @@
         consideration is whether the captured <a>display surface</a> could be somehow induced to
         present information that would otherwise be secret from the application that is receiving
         the resulting media.
+      </p>
+      <p>
+        When rendering applications fullscreen, an user agent must be 
+        prepared to handle certain transient behaviors where the action to capture an 
+        application/window in fullscreen may be reverted, say due to user canceling the 
+        action immediately. In these scenarios, the user agent MUST ensure that the 
+        captured display surface is restored to the application/window  
+        originating the action.
       </p>
       <section>
         <h2>
@@ -469,7 +503,7 @@
         </p>
         <p>
           Two forms of consent interaction are described: <a>active user consent</a> and a range of
-          <a>elevated permissions</a>. These are non-normative recommandations only.
+          <a>elevated permissions</a>. These are non-normative recommendations only.
         </p>
         <section>
           <h2>


index.html
```diff
@@ -353,6 +353,35 @@ <h2>
               </tr>
             </tbody>
           </table>
+          <div class="note">
+            <p>Below are few non-normative recommendations when capturing display for 
+              applications that support full screen rendering.
```
martinthomson: Cut this paragraph.

index.html
```diff
@@ -353,6 +353,35 @@ <h2>
...
+            </p>
+            <p>
+              For applications that support full screen rendering, like powerpoint for example,
```
martinthomson: s/like powerpoint for example/such as presentation applications/

index.html
```diff
@@ -353,6 +353,35 @@ <h2>
...
             </tbody>
           </table>
+          <div class="note">
+            <p>Below are few non-normative recommendations when capturing display for 
+              applications that support full screen rendering.
+            </p>
+            <p>
+              For applications that support full screen rendering, like powerpoint for example, 
+              the decision of the display surface that gets captured while in full screen mode is 
+              specific to the user agent implementation. For consistency reasons,
```
martinthomson: Missing sentence.  "A user agent MAY choose to switch from capture of a window to the full screen when the associated application shows a full screen."

index.html
```diff
@@ -353,6 +353,35 @@ <h2>
               </tr>
             </tbody>
           </table>
+          <div class="note">
```
martinthomson: I think that you should make a new section and mark that as non-normative.

index.html
```diff
@@ -353,6 +353,35 @@ <h2>
...
+            <p>Below are few non-normative recommendations when capturing display for 
+              applications that support full screen rendering.
+            </p>
+            <p>
+              For applications that support full screen rendering, like powerpoint for example, 
+              the decision of the display surface that gets captured while in full screen mode is 
+              specific to the user agent implementation. For consistency reasons, 
+              an user agent must ensure that the display surface captured in full screen 
+              mode be unambigously mapped to the application/window where the action was triggered 
+              in the first place.
```
martinthomson: The point here is for it to be clear that in a multi-window application the user agent is reasonably certain that the full screen switch was triggered by the window that was being captured.  You can point out here that this might be done by ensuring that this transition is only made when the captured window has focus before the transition.

index.html
```diff
@@ -353,6 +353,35 @@ <h2>
...
+              in the first place. 
+            </p>
+            <p>
+              The choice of <code><a>DisplayCaptureSurfaceType</a></code> and 
+              <code><a data-link-for="MediaTrackConstraintSet">logicalSurface</a></code> may 
+              further influence what gets captured. For example, when an application/window 
+              is triggered to go full screen and the DisplayCaptureSurfaceType of "monitor" 
+              is selected, the choice of the "monitor" display surface selected for
+              capture is solely left to the user agent, especially when there 
+              are mutliple options available.
```
martinthomson: This example isn't particularly informative.  You might like to note that if the logical surface is being captured, then it might not make sense to follow the transition, but it might make more sense for a display surface, for which the expectation is that what is seen is captured.

index.html
```diff
@@ -353,6 +353,35 @@ <h2>
...
+            </p>
+            <p>
+             An user agent must also be prepared to handle certain transient behaviors where 
+             the action to capture an application/window in full screen may be reverted, say due
+             to user canceling the action immediately. In these scenarios, the user agent
+             should ensure that the captured display surface is restored to the application/window 
+             originating the action.
```
martinthomson: I think that you need to emphasize this in the security considerations.  Possibly with a MUST slapped on it. If the intent of the user is to share powerpoint and a flash of their entire desktop is shown when the presentation ends, that's muy bad.

martinthomson (commented):
Thanks for taking this on Suhas.  I think that you have the skeleton here, but I think that you are better off making this a new section, and making it normative overall.

That the browser-choice stuff is non-normative should be obvious, but there's a hidden normative requirement here.

vivienlacourba:
relates to #29

suhasHere:
@martinthomson updated the PR based on your review comments. Please let me know what do you think.
Thanks

index.html
```diff
@@ -379,6 +379,32 @@ <h2 id="deviceId">
           duplicate any existing values.
         </p>
       </section>
+      <section>
+        <h2 id="fullscreen">
+          Fullscreen Rendering considerations
+        </h2>
+          <p><i>This section is non-normative.</i></p>
+          <p>
+            For applications that support full screen rendering such as powerpoint applications,
```
martinthomson, line 388: s/powerpoint/presentation/

index.html
```diff
@@ -379,6 +379,32 @@ <h2 id="deviceId">
...
+            the decision of the display surface that gets captured while in fullscreen mode is 
+            specific to the user agent implementation. A user agent MAY choose to switch from 
+            capture of a window to the fullscreen when the associated application goes 
+            fullscreen. For consistency reasons, an user agent must ensure that the display surface 
+            captured in fullscreen mode be unambiguously mapped to the application/window where 
+            the action was triggered in the first place. This might be achieved by ensuring that
```
martinthomson, line 394: I think that a rewording is necessary.  The point here is that the browser needs to be certain that the fullscreen view is attributed to the application or window that is being shared.  It cannot display a fullscreen view for another window in the same application, for instance.

index.html
```diff
@@ -379,6 +379,32 @@ <h2 id="deviceId">
...
+          </p>
+          <p>
+            The choice of <code><a>DisplayCaptureSurfaceType</a></code> and 
+            <code><a data-link-for="MediaTrackConstraintSet">logicalSurface</a></code> setting 
+            may further influence what gets captured. For example, when an application/window 
+            is triggered to go fullscreen with the
+            <code><a data-link-for="MediaTrackConstraintSet">logicalSurface</a></code> value 
+            of false and the DisplayCaptureSurfaceType of "monitor", the choice of the "monitor" 
+            <a>visible display surface</a> selected for capture is solely left to the user agent, 
+            especially when there are multiple options available.
```
martinthomson, line 405: This example isn't working for me now.  Not sure what it is trying to say (logicalSurface doesn't really make sense for monitor).

index.html
```diff
@@ -418,6 +444,14 @@ <h2>
         present information that would otherwise be secret from the application that is receiving
         the resulting media.
       </p>
+      <p>
+        When rendering applications fullscreen, an user agent must be 
+        prepared to handle certain transient behaviors where the action to capture an 
+        application/window in fullscreen may be reverted, say due to user canceling the 
+        action immediately. In these scenarios, the user agent MUST ensure that the 
+        captured display surface is restored to the application/window  
+        originating the action.
```
martinthomson, line 453: add "without showing the contents of other applications or windows." to the last sentence.

martinthomson (changes requested):

index.html
suhasHere: @martinthomson I thought the above meant to say so with "must ensure that the display surface  captured in fullscreen mode be unambiguously mapped to the application/window where  the action was triggered in the first place" 

Should we reword this to be more lucid ?

index.html
suhasHere: @martinthomson The above talks about logicSurface being set to False to imply visible display surface. so the visible display surface area being picked is left to UA when there are multiple monitors. 

Should we just not use the word logicSurface and instead say about <a>visible display surface</a> directly?

index.html
suhasHere: Argh !! Fixed it in my local copy. Thanks for catching it.

index.html
martinthomson: I think that the tautology in the example is the problem.  I think that a better example might be "For example, a window that is shared might effectively capture a complete monitor when it transitions to full screen; when exiting full screen, the user agent needs to ensure that what is displayed on the monitor after the transition is not captured."  The use of logicalSurface and DisplayCaptureSurfaceType is a little distracting.

index.html
martinthomson: Yeah, I think that this is a little obtuse.  The benefit of coming back to this with fresh eyes.  

> When following a transition to full screen, a user agent must be certain that what is displayed in full screen can be attributed to the application or window that is being captured.

Also, I didn't pick up on this, but it's "a" user agent.

suhasHere: Thanks .. Agree this is more cleaner :-)

index.html
suhasHere: @martinthomson The idea we were trying to say here is : if there are more than one monitors available and the application is spread across them and a capture is called --> the choice of which visible display surface area (the specific monitor)  to capture is UA specific" ..  

I agree the tautology usage in the PR is not helping with the intent. If you agree with the intent here, I can reword the PR to talk about multiple monitor scenario to be more clear.

index.html
```diff
@@ -379,6 +379,32 @@ <h2 id="deviceId">
...
+        <h2 id="fullscreen">
+          Fullscreen Rendering considerations
+        </h2>
+          <p><i>This section is non-normative.</i></p>
+          <p>
+            For applications that support full screen rendering such as powerpoint applications,
+            the decision of the display surface that gets captured while in fullscreen mode is 
+            specific to the user agent implementation. A user agent MAY choose to switch from 
+            capture of a window to the fullscreen when the associated application goes 
+            fullscreen. For consistency reasons, an user agent must ensure that the display surface
```
henbos, line 392: Why be ambiguous as to what happens? I would imagine sharing a window that goes into fullscreen would be the same as if that window got maximized from the JS application's perspective.

My concern with "switching mode" is if this accidentally has other implications due to any differences in modes, if any other part of the spec says that anything should be done differently depending on mode.

If we switch to "fullscreen mode" maybe we accidentally leak information about other windows that shouldn't be shared, like OS UI or other windows that might cover the presentation. Or maybe this wording implicitly changes what audio to share, if we decide to support it.

henbos:
The "fullscreen mode" for powerpoint probably counts as a window that you can select.

index.html
martinthomson: > I would imagine sharing a window that goes into fullscreen would be the same as if that window got maximized from the JS application's perspective.

It isn't from the perspective of the platform (at least in some cases).  The fullscreen view can be an entirely new window, rather than a resizing of the shared one.  Getting that right is tricky.

henbos:
We need to verify:
- Does an application report "I am going into fullscreen mode now", and should that be handled as a "window resize"?
- If the "fullscreen mode" means spawning new windows, we should probably not share that, rather the user should share the monitor?

martinthomson:
I'm told that Powerpoint creates a new window, and that we want that window to be shared.

henbos:
From an implementation perspective, can we tell the difference between an application spawning a new window (that shouldn't be shared), and an application spawning a new window that we should switch focus to?

If "fullscreen mode" is something that only one window can have at a time, then knowing that the fullscreen window that was just created belong to the same application as the originally shared window _may_ be enough to conclude we should switch which window is shared (if there are even OS APIs for the user agent to know this)... If "fullscreen mode" is just a borderless maximized window, with presentation notes still visible to the presenter (but not in the "fullscreen window"), then I'm not sure there's any way to tell the difference between a new window being spawned and an application wanting to do fullscreen.

Unless we can ensure that the user agent really can tell the difference and share the right window, I would assume that the way to achieve this is perhaps one of the following...
- The user shares the monitor that the fullscreen window will occupy, rather than sharing the original powerpoint window. This may share more than is desired.
- We allow the user agent to switch which window is shared on the fly. This allows the user agent to implement a picker for the end-user can manually switch which window to share to the fullscreen window. This adds user agent and user action complexity, and may only work with borderless fullscreen? A true fullscreen window would steal focus from any UI picker that the user agent presents. It also has problems with what to do when exiting fullscreen - if the shared window is destroyed, how do we know that we should switch back to the original window that is no longer being shared?

alvestrand:
Closing PR for now. We'll ask for another discussion of #29   at TPAC.

PR #58, opened 2018-04-12
Issue 38 - Cursor constraint

suhasHere:
Following up on the discussions in the issue, this PR adds cursor constraint that enumerates the conditions for capturing the cursor.

### DIFF
--- a/index.html
+++ b/index.html
@@ -245,8 +245,8 @@
           New Constraints for Captured Display Surfaces
         </h2>
         <p>
-          Following new constraints are defined that allow an application to observe properties of the
-          selected <a>display surface</a>. Since the source of media cannot be changed after a
+          Following new constraints are defined that allow an application to observe the properties 
+          of the selected <a>display surface</a>. Since the source of media cannot be changed after a
           <code>MediaStreamTrack</code> has been returned and constraints do not affect the
           selection of <a>display surfaces</a>, these constraints cannot be changed by an
           application.
@@ -414,7 +414,9 @@
                 </td>
                 <td>
                   a <a>motion</a> cursor capture constraint includes the cursor in the captured display 
-                  surface when the cursor/pointer is moved.
+                  surface when the cursor/pointer is moved. The captured cursor is removed when there 
+                  is no further movement of the pointer/cursor for certain period of time, as determined 
+                  by the <a>user agent</a>.
                 </td>
               </tr>
             </tbody>
--- a/index.html
+++ b/index.html
@@ -245,7 +245,7 @@
           New Constraints for Captured Display Surfaces
         </h2>
         <p>
-          Two constraints are defined that allow an application to observe properties of the
+          Following new constraints are defined that allow an application to observe properties of the
           selected <a>display surface</a>. Since the source of media cannot be changed after a
           <code>MediaStreamTrack</code> has been returned and constraints do not affect the
           selection of <a>display surfaces</a>, these constraints cannot be changed by an
@@ -261,6 +261,11 @@
           allows an application to observe whether the surface that is captured is a <a>logical
           display surface</a>, rather than the <a>visible display surface</a>.
         </p>
+        <p>
+          The <code><a data-link-for="MediaTrackConstraintSet">cursor</a></code> constraint
+          allows an application to specify if the cursor should be included in the captured 
+          display.
+        </p>
         <section>
           <h2>
             Extensions to <a class="idlType">MediaTrackConstraintSet</a>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -245,7 +245,7 @@ <h2 id="display-constraints">
           New Constraints for Captured Display Surfaces
         </h2>
         <p>
-          Two constraints are defined that allow an application to observe properties of the
+          Following new constraints are defined that allow an application to observe properties of the
```
martinthomson: Missing "the"

martinthomson (approved):
Probably better to change the authors in a separate PR, but this looks fine.

Do you think that there is any point in mentioning that the browser needs to determine how long to render the cursor for after movement stops?

suhasHere:
thanks @martinthomson for the review. 

Sure adding a statement to remove the cursor when there is no movement might be handy. Would something like this work ?

"The captured cursor is removed when there is no further movement of the pointer for certain period of time (as determined by the user agent) to ensure smoother user experience."

martinthomson:
Yeah, that works.  I would drop the tail: "The captured cursor is removed when there is no further movement of the pointer for certain period of time, as determined by the `<a>`user agent`</a>`."

index.html
```diff
@@ -354,6 +370,58 @@ <h2>
...
+                <td>
+                  a <a>never</a> cursor capture constraint omits the cursor from the captured display surface.
+                </td>
+              </tr>
+              <tr>
+                <td>
+                  <dfn><code id="idl-def-CursorCaptureConstraint.always">always</code></dfn>
+                </td>
+                <td>
+                  a <a>always</a> cursor capture constraint includes the cursor in the captured display surface.
```
juberti, line 407: a -> an

index.html
```diff
@@ -245,8 +245,8 @@ <h2 id="display-constraints">
           New Constraints for Captured Display Surfaces
         </h2>
         <p>
-          Two constraints are defined that allow an application to observe properties of the
-          selected <a>display surface</a>. Since the source of media cannot be changed after a
+          Following new constraints are defined that allow an application to observe the properties 
+          of the selected <a>display surface</a>. Since the source of media cannot be changed after a
```
juberti, line 249: This sentence seems wrong - we do want the app to change the constraints, not just observe them.

juberti:
Think this one still needs a bit more work.

index.html
martinthomson: Good catch.

@suhasHere, we need to reframe this section.  Rather than do the obvious, and split the section on constraints into two, we should make it clear that though constraints don't affect selection (as much as people might like them to), they still affect the shape of the stream that is produced.

suhasHere: @juberti @martinthomson I was planning to do constraints related work in a separate PR that also deals with bringing back few constraints as decided in the last meeting. Would it be ok, if I address the general constraints ideas in there ?

index.html
suhasHere: @juberti my bad .. I will get this fixed.

guest271314:
@juberti FYI Tried `.getDisplayMedia({video: {cursor: "never"}})`, `videoTrack.applyConstraints({cursor: "never"})` and `videoTrack.cursor = "never"` at Chromium 72. Neither removed the cursor entirely from the captured media stream

cursor

cursor_1


After setting the constraints `videoTrack.getSettings()`

https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/getConstraints#Return_value

> To get the currently active settings for all constrainable properties, you should instead call `getSettings()`.

the resulting object still had `cursor` set to `"always"`

```
{aspectRatio: <value>
cursor: "always" // after applying constraints to the (video) `MediaStreamTrack`
deviceId: <value>
displaySurface: <value>
frameRate: <value>
height: <value>
logicalSurface: <value>
resizeMode: <value>
videoKind: <value>
width: <value>}
```

Outside of using `.pointerLock()` CSS `element {cursor:none}` was the code that resulted in the cursor not being rendered as part of the media stream.

The screen-capture directory at wpt there does not appear to include a test for cursor.

guest271314:
https://bugs.chromium.org/p/chromium/issues/detail?id=959651

CapOM:
Hi, can you confirm that when a capture is started through `.getDisplayMedia({video: {cursor: "always"}})` then later on a call to  `videoTrack.applyConstraints({cursor: "never"})` should make the cursor disappear on the fly, without the need to re-select the source ? I.e. further calls to `never` and `always` on the track `must` be supported to allow developers to show and hide cursor when needed without requiring to restart the capture pipeline.
Refs: https://w3c.github.io/mediacapture-main/#dfn-applyconstraints-algorithm and https://w3c.github.io/mediacapture-main/#dfn-selectsettings .
It looks to me that the spec allows that but it is up to the implementation to decide whether to succeed or fail.
I wonder if it would be possible to make it a requirement or at least have a way to query if the implementation supports that.
Thx!

Issue #31, opened 2015-10-30
define behaviors of all existing constraints that should apply in the screen capture


ShijunS:
At the TPAC session, we decided to clarify the meaning of all the existing constraints that can be applicable to the screen capture.  For example, resolution, framerate, and maybe "exact", etc.  It'd be good to go through the constraint model in the current Media Capture spec so we don't miss anything.

martinthomson:
The document says:

> getDisplayMedia supports all constraints that are defined for use with getUserMedia, however it does not use these constraints to select a specific source or to narrow the set of options for source selection (see 5.2 Constraining Display Surface Selection).

@ShijunS, is this enough for you?

ShijunS:
I'd suggest honoring the frameRate constraint, then ignoring all others.

martinthomson:
I looked through the list of constraints here: https://w3c.github.io/mediacapture-main/getusermedia.html#idl-def-MediaTrackSupportedConstraints

I couldn't see a reason not to support all of these, apart from those that a physically impossible (facingMode, deviceId) and those fit into the existing rules for unsupportable constraints.

ShijunS:
Clearly groupId and sourceType are not applicable either.  

We discussed the resolution topic at TPAC, my recollection is that we don't want to enforce any resolution constraints (e.g. width, height, aspectRatio).  It's up to the end user to pick the display surface.  Meanwhile, it's desirable to keep the captured surface data with the original resolution.  Otherwise, the spec will have to define the constraints in details, which I would not recommend before we get further experience based on implementations as well as feedback from web developers.

martinthomson:
There are two applications of constraints: selection of a source, and shaping of the `MediaStreamTrack`.  The first is out of scope, but the second set is fine.  I think that being able to downscale screen captures will be critical (see also #35).

I don't want to create a situation where we add a "color" (or "colour") constraint that controls the colour space of a captured track and have it unusable in this context without a spec update.

ShijunS:
The current resolution constraints (again, height, width, and aspectRatio) make sense from the webcam perspective given it is the user agent picks the resolution from various native resolutions supported by the webcam devices.  When end user picks the screen or screen region, if the spec would like to have the user agent follow the current resolution constraints, the spec would have to define cropping, scaling, and/or padding, etc. clearly.  

[...]

PR #73, opened 2018-08-27
Separate getDisplayMedia() algorithm; reintroduce constraints.

jan-ivar:
Fix for https://github.com/w3c/mediacapture-screen-share/issues/49.

### DIFF
--- a/index.html
+++ b/index.html
@@ -631,7 +631,7 @@
         </p>
         <div class="note">
           <p>
-          This exposed device identifier is not to be confused with the stable
+          This exposed deviceId identifier is not to be confused with the stable
           and private id of the same name used in algorithms to implement privacy
           indicators.
           </p>
--- a/index.html
+++ b/index.html
@@ -327,7 +327,7 @@
                     a display device, with a PermissionDescriptor named
                     <code>"display"</code>, resulting in a set of
                     provided media.</p>
-                    <p>The provided media MUST include precisely one trackß of
+                    <p>The provided media MUST include precisely one track of
                     each media type in <var>requestedMediaTypes</var>.
                     The devices chosen MUST be the ones determined by the user.
                     Once selected, the source of a
--- a/index.html
+++ b/index.html
@@ -137,6 +137,10 @@
         in whole or part, so the <a>visible display surface</a> is a strict subset of the
         <a>logical display surface</a>.
       </p>
+      <p>The terms <dfn>permission</dfn>, <dfn>retrieve the permission state</dfn>,
+      <dfn id="prompt-the-user-to-choose">prompt the user to choose</dfn>, and
+      <dfn id="create-permission">create a permission storage entry</dfn> are
+      defined in [[!permissions]].</p>
     </section>
     <section>
       <h2>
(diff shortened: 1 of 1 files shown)


jan-ivar:
Some explanation of this PR:

After deliberating with other editors, I gave up trying to preserve the call to the `getUserMedia` algorithm. A growing number of differences made it hard to deduce exactly what to implement. The lower-level algorithm calls are intact.

I also added Privacy Indicator Requirements, since these were only implied before, and with the explicit call to `getUserMedia` gone, I wanted to make sure we didn't lose them.

Finally, the content changes:
 * constraints are supported by `getDisplayMedia` again, with these caveats:
    * Using `advanced`, `min` or `exact` constraints gives `TypeError`.
    * Added note on reason for allowing `max` constraint.
 * Use the right permission algorithm (**prompt the user to choose**)
 * Disallow persisting "granted" permissions only. 
 * Proposes "screen" as permission descriptor name (do we prefer "display"?)

index.html
```diff
@@ -448,6 +630,48 @@ <h2 id="deviceId">
...
+        "https://w3c.github.io/mediacapture-main/#privacy-indicator-requirements">
+        Privacy Indicator Requirements</a> of
+      <code><a>getUserMedia</a></code>.</p>
+      <div class="note">
+        <p>
+        Since this specification forbids user agents from persisting "granted"
+        permissions, only the "Live" indicators are significant.
+        </p>
+      </div>
+      <p>The User Agent MUST NOT implement the <code><a href=
```
alvestrand: "MUST NOT implement" sounds a bit strange. Would it be clearer to say "MUST NOT fire the devicechange event based on changes in the set of available sources for getDisplayMedia"?

index.html
```diff
@@ -194,49 +198,227 @@ <h2>
...
+              </li>
+              <li>
+                <p>Let <var>originIdentifier</var> be the <a>current settings
+                  object</a>'s <a data-cite=
+                    "!HTML52/webappapis.html#responsible-browsing-context">responsible
+                    browsing context</a>'s [[!HTML52]] <a data-cite=
+                      "!HTML52/browsers.html#top-level-browsing-context">top-level
+                      browsing context</a>'s <a data-cite=
+                        "!HTML52/browsers.html#active-document">active document</a>'s
+                      origin.</p>
```
martinthomson, line 284: Wow, what a mouthful.  It's correct, but do we not have a shorthand for this we can use instead?

index.html
```diff
@@ -194,49 +198,227 @@ <h2>
...
+              </li>
+              <li>
+                <p>If the <a>current settings object</a>'s origin is
+                different from <var>originIdentifier</var>, set
+                <var>originIdentifier</var> to the result of combining
+                <var>originIdentifier</var> and the <a>current settings
+                  object</a>'s origin.</p>
```
martinthomson, line 291: I don't know what this means.  (If this is a process defined in Feature Policy, then) you need a citation.

index.html
```diff
@@ -194,49 +198,227 @@ <h2>
...
+                    limitations, jump to the step labeled <em>Permission
+                      Failure</em> below.</p>
+                  </li>
+                  <li>
+                    <p>For the origin identified by
+                    <var>originIdentifier</var>, <a>prompt the user to choose</a>
+                    a display device, with a PermissionDescriptor with its name
+                    member set to the permission name associated with
+                    <var>kind</var> for obtaining display devices (e.g. "screen"
+                    for video), resulting in a set of provided media.</p>
```
martinthomson: I think that "display" is better, and that we don't want to distinguish between video and audio capture in this context.

That's just my intuition here, so I'm open to arguments that suggest otherwise.

index.html
```diff
@@ -194,49 +198,227 @@ <h2>
...
+                    <p>The provided media MUST include precisely one track of
+                    each media type in <var>requestedMediaTypes</var>.
+                    The devices chosen MUST be the ones determined by the user.
+                    Once selected, the source of a
+                    <code><a>MediaStreamTrack</a></code> MUST NOT change.</p>
+                    <p>User Agents are encouraged to warn users against sharing
+                    <a>browser</a> display devices, or otherwise try to
+                    steer users away from sharing them.</p>
```
martinthomson: "or otherwise try to discourage their selection on the basis that these represent a significantly higher risk when shared"

index.html
```diff
@@ -194,49 +198,227 @@ <h2>
...
+                    <p>If the result of the request is "granted", then for
+                    each device that is sourcing the provided media, using
+                    the device's deviceId, <var>deviceId</var>, set
+                    [[\devicesLiveMap]]<var>[deviceId]</var> to
+                    <code>true</code>, if it isn’t already <code>true</code>,
+                    and set the
+                    [[\devicesAccessibleMap]]<var>[deviceId]</var> to
+                    <code>true</code>, if it isn’t already
+                    <code>true</code>.</p>
```
martinthomson, line 348: We need to establish what *`deviceId`* is here.  My suggestion is that while it is valuable to interact with the gUM model in this way, you want to assign a unique *`deviceId`* every time a capture is started here, only reusing the identifier while the same source is active.  That needs to be part of the preceding steps of the algorithm here.

index.html
```diff
@@ -194,49 +198,227 @@ <h2>
...
+          </dd>
+          <div class="note">
             <p>
-              If the <code>constraints</code> argument is omitted, a default value containing a
-              single <code>video</code> attribute set to <code>true</code> is assumed.
+            Min and exact constraints are disallowed by getDisplayMedia().
+            The max constraint type lets a web application provide a maximum envelope
+            for constrainable properties like width and height, should the
+            end-user resize a <a>window</a> or <a>browser</a> surface while it
+            is being captured.
```
martinthomson, line 410: This is a reasonable conclusion, but the choice on min/max doesn't seem particularly principled.  Are there other constraints that might need similar treatment?

index.html
martinthomson: What Harald said.

martinthomson (commented):
This looks pretty good.  I think that we should settle on a single permission descriptor rather than two.  I suggest "display", if only to match the API name.  How browsers manage audio capture will be tricky enough in UX without having to worry about conjoined requests being hampered by permission discrepancies.

index.html
jan-ivar: That's existing prose copied from [getUserMedia](https://w3c.github.io/mediacapture-main/getusermedia.html#constrainable-interface).

index.html
jan-ivar: Again, copied from [getUserMedia](https://w3c.github.io/mediacapture-main/getusermedia.html#constrainable-interface).

index.html
jan-ivar: This per-device permission stuff—which no-one implements btw—has already received (legitimate) flak for being overly complicated, so I worry about adding complexity to this. Since the deviceIds are never exposed, can we perhaps use the raw internal ids which are stable?

martinthomson: If the point here is to support indicators, then sure.

index.html
martinthomson: That doesn't make it good.  It just means that you now have two incomprehensible statements.

index.html
jan-ivar: Note that only `max` is allowed. `min` and `exact` are disallowed.

There's no selection, only downscaling/decimating. We never upscale, so no `min` needed.

Eliminating `min` also avoids us having to worry about `OverconstrainedError` after selection.

I don't anticipate constraints other than `width` and `height` needing this.
I don't anticipate needing other constraints types than `min`, given this repurposing.

index.html
jan-ivar: @martinthomson As such it makes this a non-change as far as this PR is concerned. Can you open an issue on [mediacapture-main](https://github.com/w3c/mediacapture-main/issues), and can we take care of it that way? It 

Just trying to unblock this PR on feature policy and https://github.com/w3c/mediacapture-main/issues/434.

jan-ivar:
@martinthomson @alvestrand Review feedback incorporated PTAL.

index.html
aboba: You might explicitly state what constraints are allowed/disallowed.

jan-ivar: @aboba I expect https://github.com/w3c/mediacapture-screen-share/issues/31 will address that.

index.html
```diff
@@ -194,49 +198,228 @@ <h2>
...
+            <p>Prompts the user for permission to live-capture their display.</p>
             <p>
-              This method operates identically to <code><a>getUserMedia</a></code>, except that it
-              acquires media from display devices, and the <code>constraints</code> argument does
-              not accept <code><a href=
-              "https://w3c.github.io/mediacapture-main/#idl-def-mediatrackconstraints">MediaTrackConstraints</a></code>
-              values. This prevents an application from influencing the selection of sources, see
-              <a href="#constraints"></a> for details.
+              This method is similar to <code><a>getUserMedia</a></code>, except that it
+              acquires media from one display device chosen by the end-user each time.
```
henbos, line 204: "one display device" sounds like single-monitor only?

alvestrand: We might add a definition of "a display device", something like "In this context, a display device consists of all the display surfaces that are presented to the user to choose from."

jan-ivar: I think this is covered by *"Multiple monitors might also be aggregated into as a single logical monitor."*

index.html
```diff
@@ -194,49 +198,228 @@ <h2>
...
+                    <a>rejected</a> with a newly
+                    <a data-link-for="exception" data-lt="create">created</a>
+                    <code>TypeError</code>.</p>
+                  </li>
+                </ol>
+              </li>
+              <li>
+                <p>Let <var>requestedMediaTypes</var> be the set of media
+                types in <var>constraints</var> with either a dictionary
+                value or a value of <code>true</code>.</p>
```
henbos, line 251: "the set of media types in constraints" sounds like a specific member inside of constraints, can this be clarified? I think its meant to say that requestedMediaTypes is the set of the same members as in constraints but only those whose value is either a dictionary or the value true.

index.html
```diff
@@ -194,49 +198,228 @@ <h2>
...
+              </li>
+              <li>
+                <p>If <var>requestedMediaTypes</var> is the empty set, set
+                <var>requestedMediaTypes</var> to a set containing
+                <code>"video"</code>.</p>
```
henbos, line 256: Should this be "video:true"?

index.html
```diff
@@ -194,49 +198,228 @@ <h2>
...
+                  object</a>'s origin.</p>
+              </li>
+              <li>
+                <p>Let <var>p</var> be a new promise.</p>
+              </li>
+              <li>
+                <p>Run the following steps in parallel:</p>
+                <ol>
+                  <li>
+                    <p>For each media type <var>T</var> in
```
henbos, line 300: Is "media type" well defined? Should it be <a></a> or a link, or somewhere clarified that this is "audio" or "video"?

index.html
```diff
@@ -194,49 +198,228 @@ <h2>
...
+                    user preference, for security reasons, or due to platform
+                    limitations, jump to the step labeled <em>Permission
+                      Failure</em> below.</p>
+                  </li>
+                  <li>
+                    <p>For the origin identified by
+                    <var>originIdentifier</var>, <a>prompt the user to choose</a>
+                    a display device, with a PermissionDescriptor named
+                    <code>"display"</code>, resulting in a set of
+                    provided media.</p>
```
henbos, line 329: 'with a PermissionDescriptor named "display",'
I'm sure this was copied from GUM, but I don't know what this means, should we file an issue?

index.html
```diff
@@ -194,49 +198,228 @@ <h2>
...
+                    <p>The provided media MUST include precisely one track of
+                    each media type in <var>requestedMediaTypes</var>.
+                    The devices chosen MUST be the ones determined by the user.
+                    Once selected, the source of a
+                    <code><a>MediaStreamTrack</a></code> MUST NOT change.</p>
```
henbos, line 334: This sentence may be up for interpretation and debatable. You may share an application, whose set of windows changes or resizes, or where the source is N/A during the time something is minimized, or we may add language about what to do if the set of monitors changes. Or something.

This is OK but we may need to revisit it.

index.html
henbos: I think this needs clarification too.

index.html
```diff
@@ -194,49 +198,228 @@ <h2>
...
+                  </li>
+                  <li>
+                    <p>For the origin identified by
+                    <var>originIdentifier</var>, <a>prompt the user to choose</a>
+                    a display device, with a PermissionDescriptor named
+                    <code>"display"</code>, resulting in a set of
+                    provided media.</p>
+                    <p>The provided media MUST include precisely one track of
+                    each media type in <var>requestedMediaTypes</var>.
+                    The devices chosen MUST be the ones determined by the user.
```
henbos, line 332: nit: Language like "devices" should be updated, we're talking about display surfaces.

index.html
henbos: nit: '... or in "rejected".'?

index.html
```diff
@@ -194,49 +198,228 @@ <h2>
...
+                    <code>true</code>, if it isn’t already <code>true</code>,
+                    and set the
+                    [[\devicesAccessibleMap]]<var>[deviceId]</var> to
+                    <code>true</code>, if it isn’t already
+                    <code>true</code>.</p>
+                    <p>The User Agent MUST NOT
+                    <a href="#create-permission">create a permission storage entry</a>
+                    with a value of "granted".
+                    </p>
+                    <p>If the result is "denied", jump to the step labeled
```
henbos, line 353: nit: I'm not sure its clear what the possible results are.

index.html
```diff
@@ -194,49 +198,228 @@ <h2>
...
+                    <p>Let <var>stream</var> be the
+                    <code><a>MediaStream</a></code> object for which the user
+                    granted permission.</p>
+                  </li>
+                  <li>
+                    <p>Run the <a href=
+                    "https://w3c.github.io/mediacapture-main/#dfn-applyconstraints-algorithm">
+                    <dfn>ApplyConstraints algorithm</dfn></a> on all
+                    tracks in <var>stream</var> with the appropriate
+                    constraints. Should this fail, let <var>failedConstraint</var>
```
henbos, line 378: nit: "with the appropriate constraints" should be clarified. As-is, both boolean and constraint set values are allowed (do we need to convert "true" to "{}"?), and it is not explicit about applying "audio" constraints to audio and "video" constraints to video. This may seem obvious but as an algorithm step it is vague.

index.html
```diff
@@ -194,49 +198,228 @@ <h2>
...
+                  <li>
+                    <p>If <var>value</var> contains a member named <code>advanced</code>,
+                    return a promise <a>rejected</a> with a newly
+                    <a data-link-for="exception" data-lt="create">created</a>
+                    <code>TypeError</code>.</p>
+                  </li>
+                  <li>
+                    <p>If <var>value</var> contains a member which in turn
+                    is a dictionary containing a member named either
+                    <code>min</code> or <code>exact</code>, return a promise
```
henbos, line 241: (If it turns out that there are meaningful differences between exact and ideal in the case of boolean constraints we can revisit this separately.)

henbos (approved):
I put down several comments, but I think they can be addressed separately. This PR is blocking other stuff, and most of the nits are inherited from GUM.

henbos:
Comments to be addressed separately.

index.html
jan-ivar: @henbos Copied from mediacapture-main.

"media type" seems defined [by extension](https://w3c.github.io/mediacapture-main/getusermedia.html#defining-a-new-media-type-beyond-the-existing-audio-and-video-types) to be audio or video, referring here to the key names "audio" or "video" in e.g. constraints like `{audio: true, video: {}}`.

The type of elements in the set `requestedMediaTypes` doesn't seem super-relevant as long as it distinguishes audio vs. video and future types, though it seems safe to infer string IMHO. Please open an issue on mediacapture-main if you think this isn't clear.

jan-ivar: It's really more of a template argument than a string I suppose. We do gymnastics like:

*"For each media type T in requestedMediaTypes, ... If the value of the T entry of constraints is "true""*

E.g. if unrolled, would read *"if the value of the audio entry of constraints"* and again for video.

Issue #66, opened 2018-08-24
Constraints for Captured Display Surfaces should move to MediaTrackSettings


uysalere:
The spec currently allows constraints to observe the properties of the selected display surface. These are only observable, such that they cannot be changed after returned. 

The way it is working right now, they would be better defined under getSettings() and partial dictionary should be defined on MediaTrackSettings. The aim here seem to be to allow the application to query the current settings of the object's constrainable properties, which getSettings() does exactly. Putting these as read-only exceptional cases under getConstraints() seems to overlap with what getsettings() should do.

alvestrand:
What properties of the selected display surface are you referring to?
Note that changes to the spec need to make sense after #49 is resolved.

jan-ivar:
The [constrainable pattern](https://w3c.github.io/mediacapture-main/getusermedia.html#constrainable-interface) appears to force our hand here. It has no concept of read-only constrainable properties, which I think means that if we want to humor that pattern, we must define these as *constrainable properties* (and ignore/throw `OverconstrainedError` on changes to them), for POLA.

That said, good observation that we're not extending `MediaTrackSettings`, which we should. We need to extend all 4 dictionaries mentioned in the [extensibility section](https://w3c.github.io/mediacapture-main/getusermedia.html#defining-a-new-constrainable-property).

uysalere:
> What properties of the selected display surface are you referring to?

I am referring to the new constraints defined in section 5.3. "Since the source of media cannot be changed after a MediaStreamTrack has been returned and constraints do not affect the selection of display surfaces, these constraints cannot be changed by an application." With this restriction in place, I believe it would be more appropriate to define them as "Setting" which is read-only.
https://w3c.github.io/mediacapture-main/getusermedia.html#terminology

> Note that changes to the spec need to make sense after #49 is resolved.

 The new constraints defined in section 5.3 are not allowed within those constraints.

> That said, good observation that we're not extending MediaTrackSettings, which we should. We need to extend all 4 dictionaries mentioned in the extensibility section.

I agree. Should we also mention which methods should user expect to fill these properties? i.e., getSettings(), getCapabilities() should but getConstraints() should not.

jan-ivar:
> more appropriate to define them as "Setting" which is read-only

That's not a separate thing. *"A setting refers to the immediate, current value of the source's constrainable properties"*.

I think "read-only" in this context means `OverconstrainedError`.

jan-ivar:
[...]

PR #75, opened 2018-09-04
Extend constraints to constrainable patterns

uysalere:
Fix for #66.

I didn't add an extension on MediaTrackCapabilities as ranges don't really make sense for these properties.

### DIFF
--- a/index.html
+++ b/index.html
@@ -240,13 +240,14 @@
       </section>
       <section>
         <h2 id="display-constraints">
-          New Constraints for Captured Display Surfaces
+          New Constrainable Properties for Captured Display Surfaces
         </h2>
         <p>
-          Following new constraints are defined that allow an application to observe the properties 
+          Following new <a href="https://www.w3.org/TR/mediacapture-streams/#defining-a-new-constrainable-property">
+          Constrainable Properties</a> are defined that allow an application to observe the properties
           of the selected <a>display surface</a>. Since the source of media cannot be changed after a
           <code>MediaStreamTrack</code> has been returned and constraints do not affect the
-          selection of <a>display surfaces</a>, these constraints cannot be changed by an
+          selection of <a>display surfaces</a>, these constrainable properties cannot be changed by an
           application.
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -259,13 +260,62 @@ <h2 id="display-constraints">
         </p>
         <p>
           The <code><a data-link-for="MediaTrackConstraintSet">cursor</a></code> constraint
-          allows an application to specify if the cursor should be included in the captured 
+          allows an application to observe if the cursor is be included in the captured
```
martinthomson: Why this change?  Control over the inclusion of a cursor is a property we explicitly wanted to allow.  It doesn't affect source selection, so it's safe.  This change would revert an agreed change.

martinthomson (changes requested):

index.html
uysalere: Current state of the spec does not allow constraints, which is contradicting this statement about selecting cursor. It would make sense after #73 lands though.

jan-ivar: https://github.com/w3c/mediacapture-screen-share/pull/73 has merged. I agree with @martinthomson that the cursor constraint is neither read-only nor a security concern, so we should revert this part.

index.html
```diff
@@ -259,13 +260,62 @@ <h2 id="display-constraints">
...
+            </dd>
+          </dl>
+        </section>
         <section>
           <h2>
             Extensions to <a class="idlType">MediaTrackConstraintSet</a>
           </h2>
+          <p>
+            <a href="https://www.w3.org/TR/mediacapture-streams/#idl-def-MediaTrackConstraintSet">
+            MediaTrackConstraintSet</a> is used for reading the current status of constraints.
```
jan-ivar, line 317: It is used both for reading and applying.

index.html
```diff
@@ -300,10 +350,60 @@ <h2>
...
+            <dt>
+               <dfn><code>cursor</code></dfn> of type <span class=
+               "idlMemberType"><a>DOMString</a></span>
+            </dt>
+            <dd>
+              <p>
+                Assumes values from the <code><a>CursorCaptureConstraint</a></code> enumeration that
+                determines if and when the cursor is included in the captured display surface.
+              </p>
+            </dd>
```
jan-ivar, line 406: I believe we also need extensions to [MediaTrackCapabilities](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediatrackcapabilities) for this to be complete.

jan-ivar (changes requested):
Great to see this added. A few missing bits, and reverting the cursor constraint bit.

index.html
jan-ivar: > I didn't add an extension on MediaTrackCapabilities as ranges don't really make sense for these properties.

String-based and boolean constrainable properties have "ranges" in the form of arrays containing all valid values.

henbos:
What is the use case of wanting to limit the user's choices?

index.html
aboba: Reverted.

jan-ivar:
@henbos Not sure I understand the question. Constraints in this spec do not limit the user's choice. This PR adds missing bits to correctly extend the constrainable pattern we're already using.

jan-ivar:
I think we can merge this and add the MediaTrackCapabilities in a second PR.

PR #80, opened 2018-10-03
Closed and Minimized Display Surfaces

### LINKED ISSUE #72: Describe what happens if the window is closed or the monitor is disconnected
I assume MediaStreamTrack.onmute/onended fires. Is there a GETUSERMEDA section we can reference?

There is also under 6.3 "In addition to feedback mechanisms, a means to for the user to stop any active capture is advisable." - this would be the same type of thing; we should define what the expected behavior is when this happens.

henbos:
Fixes #72

### DIFF
--- a/index.html
+++ b/index.html
@@ -409,6 +409,38 @@
         </dl>
       </section>
       <section>
+        <h2 id="hidden-display-surfaces">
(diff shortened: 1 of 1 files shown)


martinthomson:
Thanks, this looks right.

PR #84, opened 2018-10-18
List Constrainable Properties and their behavior.

jan-ivar:
Fix for https://github.com/w3c/mediacapture-screen-share/issues/31.
***
[Preview](https://rawgit.com/w3c/mediacapture-screen-share/205f541a668e275b3535f7adc12c6b706ed7eeec/index.html#constraints)

### DIFF
--- a/index.html
+++ b/index.html
@@ -549,7 +549,7 @@
         <ol>
           <li>
             <p>Update all affected constrainable properties at the same time.</p>
-            <p>If this causes an "overconstrained" situation, then user agent
+            <p>If this causes an "overconstrained" situation, then the user agent
             MUST ignore the culprit constraints for as long as they
             overconstrain. The user agent MUST NOT mute the track, and
             the user agent MUST NOT fire the
--- a/index.html
+++ b/index.html
@@ -390,18 +390,21 @@
         </p>
         <p>
           When <a>display surface</a> enters an inaccessible state that is not necessarily
-          permanent, the user agent MUST queue a task that <a>sets the muted state</a> of the
-          corresponding media track to <code>true</code>.
+          permanent, the user agent MUST queue a task that
+          <a href="https://www.w3.org/TR/mediacapture-streams#set-track-muted">sets the muted state</a>
+          of the corresponding media track to <code>true</code>.
         </p>
         <p>
           When <a>display surface</a> exits an inaccessible state and becomes accessible, the
-          user agent MUST queue a task that <a>sets the muted state</a> of the corresponding
-          media track to <code>false</code>.
+          user agent MUST queue a task that
+          <a href="https://www.w3.org/TR/mediacapture-streams#set-track-muted">sets the muted state</a>
+          of the corresponding media track to <code>false</code>.
         </p>
         <p>
           When a <a>display surface</a> enters an inaccessible state that is permanent (such as
-          the source application terminating), the user agent MUST queue a task that <a>ends</a>
-          the corresponding media track.
+          the source application terminating), the user agent MUST queue a task that
+          <a href="https://www.w3.org/TR/mediacapture-streams#track-ended">ends</a> the
+          corresponding media track.
         </p>
         <p>
           A stream that was just returned by <code><a>getDisplayMedia</a></code> MAY contain
@@ -448,7 +451,7 @@
           <tbody>
             <tr id="def-constraint-width">
               <td><dfn>width</dfn></td>
-              <td><code><a>ConstrainULong</a></code></td>
+              <td><code><a href="https://www.w3.org/TR/mediacapture-streams#dom-constrainulong">ConstrainULong</a></code></td>
               <td>The width or width range, in pixels. As a capability, max MUST
               reflect the <a>display surface</a>'s width, and min MUST reflect
               the width of the smallest aspect-preserving representation
@@ -456,7 +459,7 @@
             </tr>
             <tr id="def-constraint-height">
               <td><dfn>height</dfn></td>
-              <td><code><a>ConstrainULong</a></code></td>
+              <td><code><a href="https://www.w3.org/TR/mediacapture-streams#dom-constrainulong">ConstrainULong</a></code></td>
               <td>The height or height range, in pixels. As a capability, max
               MUST reflect the <a>display surface</a>'s height, and min MUST
               reflect the height of the smallest aspect-preserving
@@ -465,7 +468,7 @@
             </tr>
             <tr id="def-constraint-frameRate">
               <td><dfn>frameRate</dfn></td>
-              <td><code><a>ConstrainDouble</a></code></td>
+              <td><code><a href="https://www.w3.org/TR/mediacapture-streams#dom-constraindouble">ConstrainDouble</a></code></td>
               <td>The frame rate (frames per second) or frame rate range. As a
               capability, max MUST reflect the <a>display surface</a>'s frame
               rate, and min MUST reflect the lowest frame rate available through
@@ -473,7 +476,7 @@
             </tr>
             <tr id="def-constraint-aspect">
               <td><dfn>aspectRatio</dfn></td>
-              <td><code><a>ConstrainDouble</a></code></td>
+              <td><code><a href="https://www.w3.org/TR/mediacapture-streams#dom-constraindouble">ConstrainDouble</a></code></td>
               <td>The exact aspect ratio (width in pixels divided by height in
               pixels, represented as a double rounded to the tenth decimal
               place) or aspect ratio range. As a setting, represents
@@ -483,11 +486,12 @@
             </tr>
             <tr id="def-constraint-resizeMode">
               <td><dfn>resizeMode</dfn></td>
-              <td><code><a>ConstrainDOMString</a></code></td>
+              <td><code><a href="https://www.w3.org/TR/mediacapture-streams#dom-constraindomstring">ConstrainDOMString</a></code></td>
               <td>
                 This string (or each string, when a list) should be one of the
-                members of <code><a>VideoResizeModeEnum</a></code>. As a setting,
-                <code>none</code> means the <code><a>MediaStreamTrack</a></code>
+                members of
+                <code><a href="https://www.w3.org/TR/mediacapture-streams#dom-videoresizemodeenum">VideoResizeModeEnum</a></code>.
+                As a setting, <code>none</code> means the <code><a>MediaStreamTrack</a></code>
                 contains all bits needed to render the display in full detail,
                 which if <code>window.devicePixelRatio > 1</code>, means
                 <code>width</code> and <code>height</code> will be larger
@@ -502,7 +506,7 @@
             </tr>
             <tr id="def-constraint-displaySurface">
               <td><dfn>displaySurface</dfn></td>
-              <td><code><a>ConstrainDOMString</a></code></td>
+              <td><code><a href="https://www.w3.org/TR/mediacapture-streams#dom-constraindomstring">ConstrainDOMString</a></code></td>
               <td>
                 This string (or each string, when a list) should be one of the
                 members of <code><a>DisplayCaptureSurfaceType</a></code>. As a
@@ -514,7 +518,7 @@
             </tr>
             <tr id="def-constraint-logicalSurface">
               <td><dfn>logicalSurface</dfn></td>
-              <td><code><a>ConstrainBoolean</a></code></td>
+              <td><code><a href="https://www.w3.org/TR/mediacapture-streams#dom-constrainboolean">ConstrainBoolean</a></code></td>
               <td>
                 As a setting, a value of <code>true</code> indicates capture of
                 a <a>logical display surface</a>, whereas a value of
@@ -523,12 +527,12 @@
                 value MUST be the lone value present, rendering this
                 property immutable from the application viewpoint.
                 <p>
-                </p>
+                </p>s
               </td>
             </tr>
             <tr id="def-constraint-cursor">
               <td><dfn>cursor</dfn></td>
-              <td><code><a>ConstrainDOMString</a></code></td>
+              <td><code><a href="https://www.w3.org/TR/mediacapture-streams#dom-constraindomstring">ConstrainDOMString</a></code></td>
               <td>
                 This string (or each string, when a list) should be one of the
                 members of <code><a>CursorCaptureConstraint</a></code>. As a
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
-          The <code><a data-link-for="MediaTrackConstraintSet">logicalSurface</a></code> constraint
-          allows an application to observe whether the surface that is captured is a <a>logical
-          display surface</a>, rather than the <a>visible display surface</a>.
-        </p>
-        <p>
-          The <code><a data-link-for="MediaTrackConstraintSet">cursor</a></code> constraint
-          allows an application to specify if the cursor should be included in the captured
-          display.
-        </p>
+        <p>Constraints serve a different purpose for display device capture. They
```
martinthomson: Different when compared to... ?

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+        do not aid discovery, instead they are applied only after user-selection.
+        Some of these constraints enable user agent processing like downscaling
```
martinthomson: You need to add `<p>` in here.

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+              frame decimation by the user agent.</td>
+            </tr>
+            <tr id="def-constraint-aspect">
+              <td><dfn>aspectRatio</dfn></td>
+              <td><code><a>ConstrainDouble</a></code></td>
+              <td>The exact aspect ratio (width in pixels divided by height in
+              pixels, represented as a double rounded to the tenth decimal
+              place) or aspect ratio range. As a setting, represents the aspect
+              ratio of the <a>display surface</a>. As a capability, min and max
+              both MUST be the setting value, rendering this property immutable
```
martinthomson:
```suggestion
              both MUST be the current setting value, rendering this property immutable
```

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+        <p>Constraints serve a different purpose for display device capture. They
+        do not aid discovery, instead they are applied only after user-selection.
+        Some of these constraints enable user agent processing like downscaling
+        and frame decimation, as well as display-specific features. Others enable
+        observation of inherent properties of a user-selected
+        <a>display surface</a>, as capabilities and settings.</p>
+        <p>The following new and existing <code><a>MediaStreamTrack</a></code>
+        <a href="https://www.w3.org/TR/mediacapture-streams/#constrainable-properties">
+        Constrainable Properties</a> are defined to apply to the user-selected
+        video <a>display surface</a>, with the following behavior:</p>
```
martinthomson, line 442: It would probably be worthwhile pointing out that the entries here, if they don't mention semantics as a setting, can't be used as one.

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+            </tr>
+            <tr id="def-constraint-logicalSurface">
+              <td><dfn>logicalSurface</dfn></td>
+              <td><code><a>ConstrainBoolean</a></code></td>
+              <td>
+                As a setting, a value of <code>true</code> indicates capture of
+                a <a>logical display surface</a>, whereas a value of
+                <code>false</code> indicates a capture capture of a
+                <a>visible display surface</a>. As a capability, this same
+                setting value MUST be the lone value present, rendering this
```
martinthomson:
```suggestion
                value MUST be the lone value present, rendering this
```

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+            <tr id="def-constraint-aspect">
+              <td><dfn>aspectRatio</dfn></td>
+              <td><code><a>ConstrainDouble</a></code></td>
+              <td>The exact aspect ratio (width in pixels divided by height in
+              pixels, represented as a double rounded to the tenth decimal
+              place) or aspect ratio range. As a setting, represents the aspect
+              ratio of the <a>display surface</a>. As a capability, min and max
+              both MUST be the setting value, rendering this property immutable
+              from the application viewpoint. However, these values may change
+              if the end-user is able to resize the <a>display surface</a>.</td>
```
martinthomson: You can drop the last sentence here given that you have a nice note after the table.

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+        </table>
+        <p>When inherent properties of the underlying source of a user-selected
+        <a>display surface</a> change, for example in response to the end-user
+        resizing a captured window, and these changes render the capabilities
+        and/or settings of one or more constrainable properties outdated, the
+        user agent MUST queue a task to update all affected constrainable
+        properties at the same time.</p>
+        <section>
+          <h2>
+            Downscaling and frame decimation
```
martinthomson: Title Case Headings

index.html
```diff
@@ -751,21 +855,12 @@ <h2 id="deviceId">
...
-          constraint, since this would allow applications to influence selection; setting
-          <code>deviceId</code> constraint can only cause the resulting
-          <code><a>MediaStreamTrack</a></code> to become overconstrained.
-        </p>
-        <p>
-          A display capture source is represented in the <code><a>MediaStreamTrack</a></code> API
-          as having <code>deviceId</code> and <code>groupId</code> attributes that are randomized
-          each time a <code><a>MediaStreamTrack</a></code> is connected. These values cannot
-          duplicate any existing values.
+          constraint, since their deviceIds are not exposed.
```
martinthomson: use `<code>`

martinthomson (commented):
Solid improvement here.

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+            </tr>
+          </thead>
+          <tbody>
+            <tr id="def-constraint-width">
+              <td><dfn>width</dfn></td>
+              <td><code><a>ConstrainULong</a></code></td>
+              <td>The width or width range, in pixels. As a capability, max MUST
+              reflect the <a>display surface</a>'s width, and min MUST reflect
+              the width of the smallest aspect-preserving representation
+              available through downscaling by the user agent.</td>
```
henbos, line 458: As a setting: What is allowed? Min and max? Only max? Do we scale up, do we scale down?

As a capability: Downscaling because of what? Because of height constraints?

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+            <tr id="def-constraint-frameRate">
+              <td><dfn>frameRate</dfn></td>
+              <td><code><a>ConstrainDouble</a></code></td>
+              <td>The frame rate (frames per second) or frame rate range. As a
+              capability, max MUST reflect the <a>display surface</a>'s frame
+              rate, and min MUST reflect the lowest frame rate available through
+              frame decimation by the user agent.</td>
+            </tr>
+            <tr id="def-constraint-aspect">
+              <td><dfn>aspectRatio</dfn></td>
```
henbos, line 478: Can you constrain this? Do we create black bars etc to respect it, or do we throw or ignore if you try to enforce a ratio?

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+              <td><code><a>ConstrainDouble</a></code></td>
+              <td>The exact aspect ratio (width in pixels divided by height in
+              pixels, represented as a double rounded to the tenth decimal
+              place) or aspect ratio range. As a setting, represents the aspect
```
henbos: If you have to preserve the aspectRation to the tenth decimal place you might be forced to only support very standard types of resolutions or else you end up with more decimal places. Is it more practical to define this as min width divided by min height and make sure our min width and min height definitions make sense?

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+                which if <code>window.devicePixelRatio > 1</code>, means
+                <code>width</code> and <code>height</code> will be larger
+                than the display's appearance from an end-user viewpoint would
+                suggest, whereas <code>crop-and-scale</code> means the
+                <code><a>MediaStreamTrack</a></code> contains an
+                aspect-preserved representation of the <a>display surface</a>
+                that has been downscaled by the user agent, but not cropped. As
+                a capability, the values <code>none</code> and
+                <code>crop-and-scale</code> both MUST be present. In other
+                words, the user agent MUST support downscaling.
```
henbos: But not upscaling?

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+              </td>
+            </tr>
+            <tr id="def-constraint-displaySurface">
+              <td><dfn>displaySurface</dfn></td>
+              <td><code><a>ConstrainDOMString</a></code></td>
+              <td>
+                This string (or each string, when a list) should be one of the
+                members of <code><a>DisplayCaptureSurfaceType</a></code>. As a
+                setting, indicates the type of <a>display surface</a> that
+                is being captured. As a capability, the setting value MUST be
```
henbos, line 514: Do we allow this changing (just like aspectRatio can change)? I think it should be allowed by the user agent to change what display surface is being captured in real-time, e.g. based on browser UI.

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+                the lone value present, rendering this property immutable from
```
henbos, line 515: What does "the lone value present" mean? The string "present"?

index.html
```diff
@@ -456,32 +421,171 @@ <h2 id="constraints">
...
+            <p>The max constraint type lets a web application provide a maximum
+            envelope for constrainable properties like width and height. This is
+            helpful to limit extreme aspect ratios, should the end-user resize a
+            <a>window</a> or <a>browser</a> surface while it is being captured.
+            </p>
+            <p>While min and exact constraints produce TypeError on
+            getDisplayMedia(), this specification does not alter the
+            track.applyConstraints() method. Therefore, they may instead produce
+            OverconstrainedError or succeed depending on values. Since there is
+            no effect of being "overconstrained", they serve no purpose.</p>
```
henbos: Can exact behavior be specified on a per-constraint basis? Sounds like there is room for interpretation.
I find mentioning OverconstrainedError throwing followed by being "overconstrained" not having an effect/serving no purpose confusing. What are we trying to say?

index.html
jan-ivar: Actually, they don't all mention "As a setting" but that doesn't mean they won't get one, since the constrainable pattern ensures you get all aspects of a constrainable property.

I tried to follow the pattern of [4.3.8 Constrainable Properties](https://www.w3.org/TR/mediacapture-streams/#constrainable-properties) in gUM, where we rely on an understanding of the constrainable pattern.

E.g. for `width` in gUM, all it says is *"The width or width range, in pixels. As a capability, the range should span the video source's pre-set width values with min being the smallest width and max being the largest width."*

It is therefore assumed that any constraint within min and max are valid, may get applied and reflected in *getSettings()*.

Similarly, any setting *not* in the list at all, does not count in the domain.

Do you still see the need for a change? If so, which specifically?

jan-ivar: Specifically, I don't think we can say *"if it's not on this list, then ignore it"* since that would lock out any extension spec ever.

index.html
jan-ivar: As a setting: [MediaTrackSettings](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediatracksettings) contains only values, not ranges, so this should be clear.

This is mirrored on [similar language in gUM](https://www.w3.org/TR/mediacapture-streams/#constrainable-properties). Describing the capability range is meant to be sufficient to deduce the other behaviors: only settings within that range are valid, and because the original size is the max, it follows that sizes larger than the original are not allowed. No up-scaling is mentioned specifically in a separate section.

As a capability: I broke out a section dedicated to detailing the downscaling behavior. Too much for the table.

index.html
jan-ivar: If you try
```
await track.applyConstraints({aspectRatio: {max: 0.2})
```
it will fail with `OverconstrainedError`. No black bars. This is explained in the downscaling section (there is no "overconstrained" condition. Let me know if it's not sufficient.

index.html
jan-ivar: This is a good point. I'll have to find a way to explain this here and in the downscaling section.

index.html
jan-ivar: I've added a sentence that UAs MUST NOT upscale to the separate section.

index.html
jan-ivar: By "changing" if you mean though applying constraints, then no, we do not allow applications to affect `aspectRatio`. It is *"immutable from the application viewpoint"*.

The spec already explicitly says `displaySurface` is read-only, and this PR is consistent with that.

I agree with the spec here, and I disagree applications should be able to change `displaySurface` since this undermines the model as exposed to users, and would be surprising.

In my view, a user agent that wants to implement some sort of "follow my active tab" exposure, would need to implement a singular "smart" virtual device (a `logicalSurface` if you will), and be consistent within this model (meaning: make it a discrete choice in the UA selector, and preferably explain it well to the user).

Personally, I question the security of letting JS applications peruse one's tabs.

index.html
jan-ivar: As a capability, the type is `sequence<DOMString>`. I'll add a PR to expand [MediaTrackCapabilities](https://w3c.github.io/mediacapture-main/getusermedia.html#media-track-capabilities).

index.html
jan-ivar: > Can exact behavior be specified on a per-constraint basis?

We're not doing so, so the only definitions of `exact` that are relevant, are those already defined in *mediacapture-main*. In all those cases *exact* is either identical to `min` = `max` = some value, or for constraints without `min` and `max` a single value that must be met, or fail with `OverconstrainedError`.

> I find mentioning OverconstrainedError throwing followed by being "overconstrained" not having an effect/serving no purpose confusing. What are we trying to say?

I'm probably saying too many things at once. I'm trying to say `min` and `exact` are not useful. What purpose does throwing `OverconstrainedError` serve here?

As to "overconstrained" having no effect, this refers to the language about "overconstrained" above this note.

index.html
jan-ivar: I've elaborated on the "overconstrained" situation, PTAL.

jan-ivar:
@alvestrand For better or worse, since these are constraints, downscaling is effectively governed by the fitness distance algorithm. I'll mention that explicitly.

Here's a [demo](https://jsfiddle.net/jib1/8t16czgy/).

jan-ivar:
@martinthomson At TPAC we agreed we could merge this. Good to go?

jan-ivar:
ah conflicts. nm.

jan-ivar:
Should be good now.

martinthomson:
Thanks for doing this.

PR #86, opened 2018-10-30
Move getDisplayMedia to MediaDevices

youennf:
As per last TPAC F2F, we want to move getDisplayMedia to MediaDevices


martinthomson:
thanks!

Issue #39, opened 2016-05-11
non-top-level browsing contexts


martinthomson:
getUserMedia is getting a sandbox flag: https://github.com/whatwg/html/pull/1211

Do we want one too, or do we instead want to forbid the use of getDisplayMedia from sandboxed iframes entirely?

alvestrand:
to be specific - the flag has been in getusermedia for a while. After discussion with @domenic, we're moving it to HTML.

mhofman:
I don't think we should prevent gDM from sandboxed iframes, the use cases for allowing gUM are the same for allowing gDM.

Is there any reason we can't share the same flag ?

martinthomson:
I will check, but I believe that there is a chance that we will block access to this API from frames entirely.  Why?  Because it's very hard to explain to a user that "example.com" wants something when they are on "other.example.net" and the decision about trust is much more important.

ShijunS:
I'd agree with @mhofman.  We should keep consistency between gDM and gUM in terms of support from sandboxed iframes.  The potential "trust" challenge would be of no difference from that of gUM.  
It'd be an interesting question though as to whether we could share the same flag.

alvestrand:
It would make sense to treat screen capture as a separate permission under the "powerful feature access" spec, which is supposed to control access to powerful features from iframes.

I think browsers should default to not permitting access from iframes in both cases, but it is nice to give the containing iframe separate controls for the two features.

ShijunS:
@Alvestrand, agree on separate iframe controls.

mhofman:
I understand the need for separate controls, but I would like to preserve access to this API from iframes.

There are some valid use cases for them. In particular, imagine a "support" service provider that offers a solution website owners can easily add to their page to offer support capabilities for their customers.
A support solution might be implemented as a sandboxed iframe for example to ensure it doesn't interfere with the page.

IMO, with the feature disabled by default in iframes and requiring a specific flag, if an embedding website includes the flag it indicates an explicit buy in. While the prompt should probably include information about the iframe's origin, from a user point of view, the embedding website is the one that matters to give permission to.

martinthomson:
See feature-policy.

aboba:
Consensus from the April VI is to [use feature policy.](https://www.w3.org/2018/04/12-webrtc-minutes)

alvestrand:
Related to #28

aboba:
@alvestrand What are the next steps on this issue?

[...]

Issue #28, opened 2015-08-26
Interaction with permissions API


martinthomson:
We probably need a new string here: https://w3c.github.io/permissions/

Also, we need to understand how we might provide feedback to the permissions API regarding the states they expose.  It might be that we need extra information input (I think we want this) and output (I don't think that we want anything special here).

martinthomson:
We will want to surface whether the site has been "blocked" or whether a user prompt will appear.  Also, sites will want to know if the enhanced permissions required for browser sharing have been granted.  That is special.  We don't want any way to request a permission.

alvestrand:
The name "display" seems like a good one.

jan-ivar:
Fixed by https://github.com/w3c/permissions/pull/184.

PR #87, opened 2018-11-01
Define and rely on "display-capture" permission-based feature policy.

jan-ivar:
Fix for https://github.com/w3c/mediacapture-screen-share/issues/39.

### DIFF
--- a/index.html
+++ b/index.html
@@ -864,7 +864,7 @@
         >policy-controlled feature</a>
       identified by the string
       <code>"<dfn data-lt="display-capture-feature">display-capture</dfn>"</code>.
-      It's <a data-link-type="dfn"
+      Its <a data-link-type="dfn"
         href="https://wicg.github.io/feature-policy/#default-allowlist"
         >default allowlist</a> is <code>"self"</code>.
       <div class="note">
--- a/index.html
+++ b/index.html
@@ -239,32 +239,6 @@
                   <code>InvalidStateError</code>.</p>
               </li>
               <li>
-                <p>If the <a>current settings object</a>'s <a>responsible
-                  document</a> is NOT <a>allowed to use</a> the feature
-                indicated by Feature Policy [TBD],
-                return a promise <a>rejected</a> with a
-                <code><a>DOMException</a></code> object whose
-                <code><a>name</a></code> attribute has the value
-                <code>SecurityError</code>.</p>
-              </li>
-              <li>
-                <p>Let <var>originIdentifier</var> be the <a>current settings
-                  object</a>'s <a data-cite=
-                    "!HTML52/webappapis.html#responsible-browsing-context">responsible
-                    browsing context</a>'s [[!HTML52]] <a data-cite=
-                      "!HTML52/browsers.html#top-level-browsing-context">top-level
-                      browsing context</a>'s <a data-cite=
-                        "!HTML52/browsers.html#active-document">active document</a>'s
-                      origin.</p>
-              </li>
-              <li>
-                <p>If the <a>current settings object</a>'s origin is
-                different from <var>originIdentifier</var>, set
-                <var>originIdentifier</var> to the result of combining
-                <var>originIdentifier</var> and the <a>current settings
-                  object</a>'s origin.</p>
-              </li>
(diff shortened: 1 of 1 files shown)


martinthomson:
The string "display" in isolation means something different to what you intend it to mean.  I think that you want "display-capture" or something along those lines.

jan-ivar:
@martinthomson FWIW `"display"` is the name we've been discussing since https://github.com/w3c/mediacapture-screen-share/issues/28#issuecomment-417339077, and is at this point already in [permissions](https://w3c.github.io/permissions/#screen-capture) and [getDisplayMedia](https://w3c.github.io/mediacapture-screen-share/#dom-mediadevices-getdisplaymedia). How strongly do you feel about it?

index.html
```diff
@@ -296,8 +270,7 @@ <h2>
                       Failure</em> below.</p>
                   </li>
                   <li>
-                    <p>For the origin identified by
```
martinthomson, line 299: I assume that this is subsumed by the "prompt the user to choose" algorithm.

martinthomson:
I'd prefer to get the right name.  I don't want to take a name that might be used for say access to displays.

jan-ivar:
@martinthomson @alvestrand discuss. What would "access to displays" mean? We don't say "camera-capture" FWIW.

index.html
```diff
@@ -884,6 +857,34 @@ <h2 id="deviceId">
...
+          id="ref-for-concept-document①⓪">document</a>'s
+        <a data-link-type="dfn"
+          href="https://html.spec.whatwg.org/multipage/dom.html#concept-document-feature-policy"
+          id="ref-for-concept-document-feature-policy">feature policy</a>
+        determines whether any content in that document is allowed to use
+        <code>getDisplayMedia</code>. If disabled in any document, no content in
+        the document will be
+        <a data-link-type="dfn"
+          href="https://html.spec.whatwg.org/multipage/iframe-embed-object.html#allowed-to-use"
+          id="ref-for-allowed-to-use②">allowed to use</a>
```
alvestrand: Is the circled 2 on this line an accidental character?

alvestrand:
The main purpose of a display is to display stuff (facing out of the computer).
We could imagine "display" being used for "is allowed to display on the displayed surface". But it's hard to imagine a Web page without that right, so it's kind of like allowing a fish to be wet.

The permission we're mediating here is a secondary function - that of allowing the display to be scraped by the computer.

The closest permission analogy to "display" in the other direction I can think of is "speaker"; there, we mediate an outward-facing access; we never capture from a speaker.

The only reason to stick with "display" is that we're already using it, and nobody else is. I'm not greatly worried about it either way.

index.html
jan-ivar: Boiler-plate from the fullscreen spec I copied. Removed now. Thanks for spotting it!

jan-ivar:
So which do we want: `display`, `display-capture`, `screen-capture`, `display-media`?

It's a bit odd, the spec itself is titled "Screen capture", but the term is mostly used [outside this spec](https://w3c.github.io/permissions/#screen-capture). "display capture" appears unused, but would seem logical even though search engines think it a [typo](https://duckduckgo.com/?q=display+capture).

jan-ivar:
I picked one.

index.html
```diff
@@ -884,6 +857,34 @@ <h2 id="deviceId">
...
       </section>
     </section>
+    <section>
+      <h1 id=feature-policy-integration>Feature Policy Integration</h1>
+      <p>This specification defines a <a data-link-type="dfn"
+        href="https://wicg.github.io/feature-policy/#policy-controlled-feature"
+        >policy-controlled feature</a>
+      identified by the string
+      <code>"<dfn data-lt="display-capture-feature">display-capture</dfn>"</code>.
+      It's <a data-link-type="dfn"
```
martinthomson:
```suggestion
      Its <a data-link-type="dfn"
```

martinthomson (approved):

henbos:
@alvestrand to review this before merging.

alvestrand (approved):

Issue #35, opened 2015-12-11
Handle source device pixel ratio


mhofman:
Issue for discussion started on the list: https://lists.w3.org/Archives/Public/public-media-capture/2015Dec/0029.html

The spec currently doesn't mention anything about how to handle sources (screen, windows) with logical pixels that don't match physical pixels of the monitor.

The most prominent case are "retina displays", which have high density physical pixels. The OS exposes logical pixels that simulate a lower resolution screen and effectively scales things. The scaling factor is usually referred to as "device pixel ratio", at least in the Web world (See [CSS spec](https://drafts.csswg.org/cssom-view/#dom-window-devicepixelratio))

A Mac retina display with a physical size of 2880x1800 usually exposes a virtual resolution of 1440x900. The devicePixelRatio is 2.0 for such monitors.

I believe current implementations simply capture the physical pixels.

When capturing large windows or a full screen, an application might not want a size that large, especially when transmitting to another display with a 1.0 device pixel ratio (effectively zooming the content 200%).

Existing width / height constraints are not adapted to dealing with the resolution in this case:
- There is no way to know the size, physical or logical, of the source selected by the user, making fixed width/height constraints impossible to use
- Max constraints would only limit the size. The content is really meant to be scaled by 2.0, not a floating point number that would make it just fit in the constrained size
- A smaller application window might have a size below the max constraints

One workaround with the state of the current specs could be to:
- gDM
- Read width / height of stream
- use [Window.devicePixelRatio API](https://drafts.csswg.org/cssom-view/#dom-window-devicepixelratio) to guess device pixel ratio of content
- applyConstraints on stream to scale it by that ratio

Unfortunately this might fail on a variety of cases:
- On some multi-monitor systems that have different pixel ratios per monitor (Windows 10), when the browser tab and content captured are on different monitors (Especially because windows can be dragged across monitors, dynamically changing pixel ratios!)
- when the current web app content is zoomed, thus decoupling the CSS devicePixelRatio from the monitor's pixel ratio.

We can solve this by:
- Specifying that the screen share is done using logical pixels only. This might be limiting for apps that would like to capture unscaled screens. It could however be a good default.
- Providing a "constraint" on the `getDeviceMedia` request that would allow the application to indicate whether it wants a stream of logical or physical pixels.
[...]

PR #88, opened 2018-11-02
Define source pixel ratio, instead of relying on window.devicePixelRatio

jan-ivar:
Follow-up fix for https://github.com/w3c/mediacapture-screen-share/issues/35.


martinthomson:
Two changes here, but both necessary.

PR #94, opened 2018-12-05
Define restrictOwnAudio

### LINKED ISSUE #79: Constraint to exclude application audio (echo)
If I screen share in a presentation that is also a conference with remote participants, and my screen sharing includes audio (e.g. I want to show a media clip as part of the presentation) we have a risk of echo.
- If screen share contains remote participants talking, when this stream arrives at a remote participant for playout, they will hear themselves AND other remote participants twice (once because they are receiving a stream from them directly and once more because of the screen share).
- We can't just throw "echo cancellation" on the problem.

Constraints are currently not allowed to limit the user's choice, and this is generally a good thing; which sources are present are are none of the application's business. However, the application need to be able to constrain the user agent not to include audio from the application that performed the getDisplayMedia() request to avoid echo. Otherwise these applications would end up presenting the user with _false choices_. Options that, if chosen, would seemingly cause "echo bugs".

**Proposal:** `{audio:{excludeApplicationAudio:true}}` limits user agent to provide a stream that does not contain the application audio. Note that this does not say what audio the user agent must provide - the implementation/user still very much has freedom of choice - it only specifies that a particular audio source must not be present. This constrain can be fulfilled in multiple ways, e.g:
1. Exclude choices from the user. For example, "tab audio" or "window audio" are still valid choices, as long as the "tab" chosen is not the application tab. "System audio" is not a valid choice.
2. "No audio" is a valid choice. (Though in this case no audio track should be produced)
3. Manipulate the audio sources to subtract the application audio.

1 and 2 are easy to implement. 3 is likely infeasible for most platforms, but conceivable. In any case, the application should not have to care about user agent capabilities - as long as the application gets a stream that does not produce echo it is happy.

henbos:
Fixes #79.

### DIFF
--- a/index.html
+++ b/index.html
@@ -702,7 +702,7 @@
              boolean displaySurface = true;
              boolean logicalSurface = true;
              boolean cursor = true;
-             boolean excludeApplicationAudio = true;
+             boolean originAudio = true;
 };</pre>
           <dl data-link-for="MediaTrackSupportedConstraints" data-dfn-for="MediaTrackSupportedConstraints"
           class="dictionary-members">
--- a/index.html
+++ b/index.html
@@ -542,6 +542,49 @@
             </tr>
           </tbody>
         </table>
+        <p>The following new and existing <code><a>MediaStreamTrack</a></code>
+        <a href="https://www.w3.org/TR/mediacapture-streams/#constrainable-properties">
+        Constrainable Properties</a> are defined to apply to the user-selected
+        audio sources, with the following behavior:</p>
+        <table class="simple">
+          <thead>
+            <tr>
+              <th>Property Name</th>
+              <th>Type</th>
+              <th>Behavior</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr id="def-constraint-width">
+              <td><dfn>originAudio</dfn></td>
+              <td><code><a href="https://www.w3.org/TR/mediacapture-streams#dom-constrainboolean">ConstrainBoolean</a></code></td>
+              <td>
+                <p>As a setting, a value of <code>true</code> indicates that
+                audio produced by this origin is included in the captured audio,
+                a value of <code>false</code> indicates that audio from the
+                origin is not included in the captured audio.</p>
+                <p>As a constraint, this property can be constrained to the
+                value <code>false</code>, producing a source incapable of
+                including origin audio (the capability of such a source is the
+                lone value <code>false</code>). However, attempts to constrain
+                this property to the value <code>true</code> MUST result in
+                <code>OverconstrainedError</code>. That is, the application can
+                force the exclusion of the origin audio but it cannot force the
+                inclusion any particular audio source. If this property is
+                unconstrained, the resulting source MAY be capable of either the
+                value <code>true</code> or <code>false</code>.</p>
+                <div class="note">
+                  <p>Due to inherrent difficulties in isolating different audio
+                  sources on a system, the recommented solution to fulfilling
+                  <code>originAudio:false</code> is to exclude user choices. For
+                  example, a user agent that supports tab capture could fulfill
+                  this constraint by presenting the user with choices including
+                  all tabs except for the origin tab.</p>
+                </div>
+              </td>
+            </tr>
+          </tbody>
+        </table>  
         <p>When inherent properties of the underlying source of a user-selected
         <a>display surface</a> change, for example in response to the end-user
         resizing a captured window, and these changes render the capabilities
@@ -659,6 +702,7 @@
              boolean displaySurface = true;
              boolean logicalSurface = true;
              boolean cursor = true;
+             boolean excludeApplicationAudio = true;
 };</pre>
           <dl data-link-for="MediaTrackSupportedConstraints" data-dfn-for="MediaTrackSupportedConstraints"
           class="dictionary-members">
@@ -689,6 +733,15 @@
                 Whether <code><a>cursor</a></code> constraint is recognized.
               </p>
             </dd>
+            <dt>
+              <dfn><code>originAudio</code></dfn> of type <span class=
+              "idlMemberType"><a>boolean</a></span>, defaulting to <code>true</code>
+            </dt>
+            <dd>
+              <p>
+                Whether <code><a>originAudio</a></code> constraint is recognized.
+              </p>
+            </dd>
           </dl>
         </section>
         <section>
@@ -703,6 +756,7 @@
              ConstrainDOMString displaySurface;
              ConstrainBoolean   logicalSurface;
              ConstrainDOMString   cursor;
+             ConstrainBoolean   originAudio;
 };</pre>
           <dl data-link-for="MediaTrackConstraintSet" data-dfn-for="MediaTrackConstraintSet" class=
           "dictionary-members">
@@ -735,6 +789,16 @@
               <p>
                 Assumes values from the <code><a>CursorCaptureConstraint</a></code> enumeration that
                 determines if and when the cursor is included in the captured display surface.
+              </p>
+            </dd>
+            <dt>
+              <dfn><code>originAudio</code></dfn> of type <span class=
+              "idlMemberType"><a>ConstrainBoolean</a></span>
+            </dt>
+            <dd>
+              <p>
+                This constraint is only applicable to audio tracks. See
+                <code><a>originAudio</a></code>.
               </p>
             </dd>
           </dl>
--- a/index.html
+++ b/index.html
@@ -556,16 +556,16 @@
           </thead>
           <tbody>
             <tr id="def-constraint-width">
-              <td><dfn>ownAudioCancellation</dfn></td>
+              <td><dfn>restrictOwnAudio</dfn></td>
               <td><code><a href="https://www.w3.org/TR/mediacapture-streams#dom-constrainboolean">ConstrainBoolean</a></code></td>
               <td>
                 <p>As a setting, this value indicates whether or not the user
-                agent is applying <a>own audio cancellation</a> to the
+                agent is applying <a>own audio restriction</a> to the
                 source.</p>
-                <p>As a constraint, this property can be constrained to the
-                value <code>true</code>, resulting in a source that has
-                <a>own audio cancellation</a> applied.</p>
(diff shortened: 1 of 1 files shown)


henbos:
@guidou, @uysalere does this make sense to you?

uysalere:
Doesn't {audio:false} as a constraint result in the same behavior?

henbos:
Note that in any case audio is optional (per previous decisions), so audio:true does not guarantee that you get audio, but the user should be allowed to choose audio depending on UA's capabilities (or choose to only share video without having to reject the request).

{audio:false} means you don't can't get audio, audio:{originAudio:false} means you're asking for audio, and any audio is acceptable as long as it does not include that particular audio source.

Examples that would satisfy originAudio:false:

- No audio (this is always OK because audio is always optional, even in audio:true case).
- Tab capture with audio from the tab - this works for all tabs except the origin's tab, so the UA can either not allow sharing that tab or if that tab is selected you get no audio.
- Browser window, with audio from all tabs except for the origin's tab.
- All desktop capture except the origin tab's audio. I don't expect anyone to implement this due to the technical complexity, so more likely: if this constraint is present, the user cannot choose desktop capture and still get audio (exclude choice or mute stream), but if the constraint isn't present it would be perfectly valid to do desktop capture with audio if the system is capable of it.

This constraint is not about forcing the UA to include any particular audio (just like audio:true does not say what choices a UA has to be able to capture), it's about forcing the UA to exclude a particular audio source. I.e. "please don't present the user with false choices" (a false choice being one that would include the origin audio and thus, for some applications, cause echo problems).

index.html
```diff
@@ -542,6 +542,49 @@ <h2 id="constrainable-properties">
...
+          <thead>
+            <tr>
+              <th>Property Name</th>
+              <th>Type</th>
+              <th>Behavior</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr id="def-constraint-width">
+              <td><dfn>originAudio</dfn></td>
```
jan-ivar: I think we need to scope this down to an audio processing constraint, like the existing ones people are familiar with.

I also worry "origin" is confusing since several tabs may be open from the same origin, and we only mean to cancel audio from the current tab, I believe.

How about `ownAudioCancellation`? Has similar wording to `echoCancellation` which people know and understand. That way `true` would be the constraining value, which I find easier to comprehend.

See below for additional comments.

index.html
```diff
@@ -542,6 +542,49 @@ <h2 id="constrainable-properties">
...
+                inclusion any particular audio source. If this property is
+                unconstrained, the resulting source MAY be capable of either the
+                value <code>true</code> or <code>false</code>.</p>
+                <div class="note">
+                  <p>Due to inherrent difficulties in isolating different audio
+                  sources on a system, the recommented solution to fulfilling
+                  <code>originAudio:false</code> is to exclude user choices. For
+                  example, a user agent that supports tab capture could fulfill
+                  this constraint by presenting the user with choices including
+                  all tabs except for the origin tab.</p>
```
jan-ivar: This is out of scope for a constraint in *getDisplayMedia*, which doesn't even run the *applyConstraints* algorithm until *after* selection, so what this prose suggests isn't possible to implement the way this spec is written.

We've also previously decided not to allow JS to limit user selection.

The good news is I disagree we need to limit user selection over this. Refusing to share e.g. the Entire Screen because its audio would feedback, is the tail wagging the dog IMHO. We should share it as best we can without feedback or at worst without audio, when this constraint is on.

When reframed as `ownAudioCancellation` I think that'd be how users would expect it to work.

This reframing would also address how `track.applyConstraint()` should work with this new constraint. JS would expect to be able to toggle any constraint like this live and hear the difference.

Depending on the abilities of the browser and the source chosen, that difference may be either that the site's own audio is magically filtered out of the audio picture somehow, or that the audio goes silent, either all the time, or intermittently whenever audio is emitted by the site.

jan-ivar (changes requested):
My main concern is that constraints in getDisplayMedia are for processing, not selection. I've added suggestions based on that.

index.html
jan-ivar: Or we could define it as best-effort. Sites could check `track.getSettings().ownAudioCancellation == false` to see if cancellation is happening or whether they need to significantly lower the volume of participants.

henbos: I like your suggestions and think they're the way forward.

I changed it to `ownAudioCancellation`, and gave it wiggle-room so that _if_ an implementation really could cancel out the audio it is free to do so, but anyone is free to implement it as "don't include the own audio source, which worst case is no audio source". I don't think it should be "best effort" in the sense that it's optional though, if a browser can't cancel the audio it is better if it mutes the track - if the application doesn't like that, it can always applyConstraints() to not use ownAudioCancellation.

Not restricting user choices does make sense, but if the user selects audio (e.g. ticks an audio checkbox, a likely UI implementation) and the result is silence that would be more unexpected than if the browser up-front didn't allow audio. However I don't want the constraint to disallow the user from choosing audio, because then the browser wouldn't have permission to include audio even if it would have the ability to do so later.

If the track does end up muted though due to ownAudioCancellation the application would be able to detect that and inform the user that audio is muted, as well as applyConstraints() differently, so it might not be a big problem.

henbos:
PTAL @jan-ivar

index.html
jan-ivar: Good points. `restrictOwnAudio` might be another name then, if we want to imply something stronger than cancellation at times. The important point for me was that it's within the realm of a processing constraint. How UAs solve it I guess is up to them. As to limiting user choice, at least we're limiting *away* from "scary" sources in this way, so there may be some leeway here.

jan-ivar (approved):
Lgtm!

index.html
```diff
@@ -542,6 +542,42 @@ <h2 id="constrainable-properties">
...
+          <tbody>
+            <tr id="def-constraint-width">
+              <td><dfn>ownAudioCancellation</dfn></td>
+              <td><code><a href="https://www.w3.org/TR/mediacapture-streams#dom-constrainboolean">ConstrainBoolean</a></code></td>
+              <td>
+                <p>As a setting, this value indicates whether or not the user
+                agent is applying <a>own audio cancellation</a> to the
+                source.</p>
+                <p>As a constraint, this property can be constrained to the
+                value <code>true</code>, resulting in a source that has
```
henbos: nit: "when constrained" (cover both true and false)

index.html
```diff
@@ -542,6 +542,42 @@ <h2 id="constrainable-properties">
...
+          <thead>
+            <tr>
+              <th>Property Name</th>
+              <th>Type</th>
+              <th>Behavior</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr id="def-constraint-width">
+              <td><dfn>ownAudioCancellation</dfn></td>
```
henbos: restrictOwnAudio

henbos:
@henbos to merge after addressing comments

index.html
```diff
@@ -542,6 +542,42 @@ <h2 id="constrainable-properties">
...
+                <a>own audio cancellation</a> applied.</p>
+                <p>When <dfn>own audio cancellation</dfn> is applied, the user
+                agent MUST attempt to remove any audio from the audio being
+                captured that was produced by the application that performed
+                <code>getDisplayMedia</code>. If the user agent is not able to
+                remove the audio through processing it SHOULD remove the audio
+                by excluding the application's audio from being captured. If
+                this results in no audio being captured, the user agent MUST
+                keep the track muted until it is able to capture audio
+                again.</p>
```
martinthomson, line 576: "application" here is very vague.  Does it mean:

1. the browser
2. the specific browser process
3. the document origin
4. the document
5. the window/tab

Relevant questions I'd like to be able to answer:

* if capture is initiated from an iframe, does the audio from the parent get captured?
* similarly, for a parent, does the audio from an iframe get captured?
* if the same origin is playing audio from multiple tabs, is it only the tab that calls this API that is suppressed, or are all of these expected to be suppressed?

I realize that there are very few cases where audio comes from multiple sources and the primary use of this will be for origins that want to ensure that their own output, coming from the same page, is not echoed back.  But this does need to be clear.

index.html
henbos: Done

index.html
henbos: Done

index.html
henbos: I filed https://github.com/w3c/mediacapture-screen-share/issues/102

index.html
henbos: I don't see how removing the option to share audio on choices that would result in a muted track would  be a fingerprint/privacy/security issue; there is no way for the application to know whether the user de-selected audio or if the user agent did it for them. Furthermore, since audio is optional, excluding audio does not impact whether or not the getDisplayMedia() request would succeed or fail.

jan-ivar: [Not all sources are equal from a security perspective](https://blog.mozilla.org/webrtc/share-browser-windows-entire-screen-sites-trust/). The reason to limit user choice was not based on fingerprinting risk, rather the risk of letting web sites steer users toward sharing the riskier class of sources, specifically *web surfaces*, and an *entire desktop* on which a web surface is present.

Here we're ok since we're generally talking about steering people *away* from those sources. Doubly-so if we're only talking about muting audiovisual sources instead of removing them from selection entirely.

index.html
martinthomson: I should have requested changes.  It's not great to merge a PR when it is so vague, unless it is making progress on an outstanding issue.  This is adding a feature, so it could have waited.

henbos: I would normally agree, but I think it was understood by all the editors that it meant the "same JavaScript" that performed getDisplayMedia(), for example "tab" or "document" and not "browser". An "application" here is not something that is shared between tabs from the same origin, and this can be clarified. What happens in edge cases like when iframes are used, or what precise wording best nails down these nuances, should definitely be discussed, but I don't think this should block us from adding the constraint in the first place. Chrome is already looking into adding audio and I want to make sure this constraint is outlined sooner rather than later as a heads-up of the direction this is going, since the issue in question #79 is a long and complicated one, and not knowing about the possibility of such constraint might take you down wrong turns. Let's iterate on clarifying "application" at #102.

Issue #104, opened 2019-03-21
Ask for user gesture to call getDisplayMedia


youennf:
As can be seen in https://github.com/w3c/deviceorientation/issues/57, device motion/orientation access request API might require a user gesture to call this API.

Given getDisplayMedia is a powerful and privacy sensitive API, requiring a user gesture to call that API might also make sense.
This might still be possible to do so given navigator.mediaDevices.getDisplayMedia is a fairly recent API.

martinthomson:
I support this idea.  How we managed to allow websites to show popups without a user gesture is - in retrospect - surprising.  And a camera or microphone prompt is no less a popup than `window.alert()`.  Unfortunately those ones are so well baked in that clawing them back might be too disruptive.  This one though, I think that we have to have a gesture for.

/cc @johannhof, who is working on cleaning this stuff up.

PR #106, opened 2019-04-29
Add user activation check when calling getDisplayMedia

youennf:


youennf:
PR for https://github.com/w3c/mediacapture-screen-share/issues/104

martinthomson:
This is fine, though I wonder if this is not something that needs to be fixed more generically in the permissions API.

youennf:
I am not sure we can always integrate this user gesture with something like "prompt the user to choose".
It might be best from a web developer perspective that an API always/never requires a user gesture.
Also, if we had that user gesture check in place for getUserMedia, doing the check as the first step of the algorithm would decrease the possibility to use getUserMedia as a probing/fingerprinting vector.

Or are you suggesting a new "retrieve callable state" hook defined in the Permissions API?

I am fine having that discussion in Permissions API.
I would prefer moving forward with this PR in parallel.

martinthomson:
I agree.  This can progress here and then we can move to depend on whatever changes happen in permissions.

Folks at Mozilla would prefer to provide more information about permissions, such as any requirement for user activation.

jan-ivar (approved):

alvestrand:
Any comments on whether "InvalidStateError" is the right code to use?

PR #115, opened 2019-09-05
Add floor value definition and fail low max constraints early.

### LINKED ISSUE #96: Address corner-cases where getDisplayMedia() shows prompt only to fail afterwards.
Basically, it's a bug in the spec that there are corner cases where calling `getDisplayMedia` will prompt the end user needlessly, only to fail. For instance:
```js
await navigator.mediaDevices.getDisplayMedia({video: {width: {max: 0}}});
```
Fixing this is tricky as it needs to balance 3 things:
 1. **Fingerprinting** - mustn't reveal any device info through re-prodding on failure.
 2. **Interop** - Different UAs may have different downscaling limits (inherently unspec'ed by constraints)
 3. and needless prompting (this issue).

I think we should just leave room in the spec for UAs to optimize away the prompt in these edge-cases, with strict language not to take user configuration into account.

jan-ivar:
Fixes https://github.com/w3c/mediacapture-screen-share/issues/96.

### DIFF
--- a/index.html
+++ b/index.html
@@ -265,6 +265,22 @@
                     <a data-link-for="exception" data-lt="create">created</a>
                     <code>TypeError</code>.</p>
                   </li>
+                  <li>
+                    <p>If <var>CS</var> contains a member whose name,
+                    <var>failedConstraint</var> specifies a
+                    constrainable property, <var>constraint</var>, applicable to
+                    <a>display surface</a>s, and whose value in turn is a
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Odd choice in wrapping, but that is the only flaw I can see.  I shall await a review from someone present at TPAC.

jan-ivar:
Rebased.

henbos (approved):

PR #116, opened 2019-09-05
Limit pre-validation to constraints in this spec.

### LINKED ISSUE #95: Clarify that we only process/TypeError on applicable constraints
Spec says: *"For each member present in constraints whose value, value, is a dictionary, run the following steps:"*

This is too broad a pre-processing of constraints, and goes against https://github.com/w3c/mediacapture-main/pull/542.

It's also hard to implement, as WebIDL bindings typically map dictionaries to `struct` in c++, with no obvious way to enumerate dictionary members—nor should we do this, as it's a design problem that MediaTrackConstraintSet cuts across spec domains like this.

For example, screen sharing has no audio constraints (because audio constraints are *microphone* constraints). Terminology baggage.

I'll add a PR to clarify that we only throw on "applicable" constraints, which are the [ones explicitly listed in the screen share spec](https://w3c.github.io/mediacapture-screen-share/#constrainable-properties).

jan-ivar:
Fixes https://github.com/w3c/mediacapture-screen-share/issues/95.

### DIFF
--- a/index.html
+++ b/index.html
@@ -247,20 +247,21 @@
               <li>
                 <p>For each member
                 <a href="https://heycam.github.io/webidl/#dfn-present">present</a>
-                in <var>constraints</var> whose value, <var>value</var>, is a
-                dictionary, run the following steps:</p>
+                in <var>constraints</var> whose value, <var>CS</var>, is
+                a dictionary, run the following steps:</p>
                 <ol>
                   <li>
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Makes sense.  Thanks.

henbos (approved):

PR #154, opened 2020-12-08
Throw on getDisplayMedia in non-fully active documents.

jan-ivar:
Follows https://github.com/w3c/mediacapture-main/pull/491.

This used to be implied in the old user-gesture model, but got lost when we switched to [transient activation](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation) in https://github.com/w3c/mediacapture-screen-share/pull/132 which seems entirely timer-based.

### DIFF
--- a/index.html
+++ b/index.html
@@ -202,6 +202,13 @@
             steps:</p>
             <ol>
               <li>
+                <p>If the [=relevant settings object=]'s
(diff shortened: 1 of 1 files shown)


jan-ivar:
Whoops, already there.

PR #163, opened 2021-04-06
Define MediaDevices.CaptureHandle

### LINKED ISSUE #159: Allow screen-captured applications to self-declare a capture handle
## State of the Art
MediaStreamTrack.label is [specified](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/label), but the standard has very little to say about the value. More specifically, nothing is said for those MediaStreamTracks which are the result of a call to `getDisplayMedia`.

Examine the label of video tracks resulting from `getDisplayMedia` in some **current implementations**, focusing mainly on window/tab capture:
* Chrome: Exposes internal Chrome state; I've filed a bug and will be fixing it soon.
* Firefox: Does not support tab-capture, but supports window capture, and uses the window's title; if the window is a Firefox window, the label is set to the active tab's title.
* Safari: Tab/window-sharing not supported.

## Problem
The value of the label for tracks received from `getDisplayMedia` is unclear if unspecified.
Using tab titles, URL.host, etc., would be all be interesting options, but it's dangerous for browsers to unilaterally decide to capture more content than before; it's better if apps opt-in to sharing more.
(One recently opened issue that would benefit from this proposal is w3c/mediacapture-screen-share#143.)

## Suggested Solution
Let the application choose what the label should be, through a new settable attribute called `navigator.mediaDevices.preferredLabel` (or any other name). It should only be settable by the top-level document. Applications that wish to share the title can opt-in to doing so. Applications that wish to broadcast an arbitrary message to an unknown capturing app can do so. It's opt-in!

### This would let applications transmit information from captured app to the capturing app.
Yes. As does the video track itself via magic pixels, QR codes, text-and-OCR, or flashing the screen black/white to signal 0s and 1s. The capture itself is a one-way data channel. We just provide a more ergonomic side-channel to broadcast a text-based annotation.

### What if the user navigates the captured tab?
We recalculate the label. It is read-only, but it's not specified as immutable.
(We should add an event as a follow-up, IMHO. As this proposal provides sufficient value even before such an event, and even if we start out with an immutable label, I would propose not delaying until we've specified the event.)

### That lets the capturing app detect navigation.
True. That's not the only way, though. More importantly, I don't think that's a problem.

### What if preferredLabel is not set?
The user agent MAY use whichever label it wishes, as it can today.

eladalon1983:
Define a mechanism to allow an application to inform the user agent of what a meaningful label would be for the tracks that would result if the tab in which the application is running gets captured. Resolves w3c/mediacapture-screen-share#159.

***
<a href="https://pr-preview.s3.amazonaws.com/eladalon1983/mediacapture-screen-share/pull/163.html" title="Last updated on Apr 22, 2021, 12:41 PM UTC (d046db8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/163/e2a35b3...eladalon1983:d046db8.html" title="Last updated on Apr 22, 2021, 12:41 PM UTC (d046db8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -158,6 +158,7 @@
         <pre class="idl"
 >partial interface MediaDevices {
   Promise&lt;MediaStream&gt; getDisplayMedia(optional DisplayMediaStreamConstraints constraints = {});
+  attribute DOMString displayMediaCaptureHandle;
 };</pre>
         <dl data-link-for="MediaDevices" data-dfn-for="MediaDevices" class="methods">
           <dt>
@@ -365,6 +366,30 @@
                 <p>Return <var>p</var>.</p>
               </li>
             </ol>
+          </dd>
+        </dl>
+        <dl data-link-for="MediaDevices" data-dfn-for="MediaDevices">
+          <dt>
+            <dfn>displayMediaCaptureHandle</dfn>
+          </dt>
+          <dd>
+            <p>By setting this field to a non-empty string, the application
+            signals to the user agent that, should this application be captured
+            using <a>getDisplayMedia</a>, the following data may be exposed to
+            the capturing application:<p>
+            <ol>
+              <li>The captured application's origin may be exposed as
+              {{CaptureHandleType/origin}}.</li>
+              <li>The value to which <a>displayMediaCaptureHandle</a> was set
+              may be exposed as {{CaptureHandleType/handle}}.</li>
+            </ol>
+            <p>Setting the value to a new non-empty value updates handle.</p>
+            <p>Setting the value to the empty string rescinds the permission to
+            expose either {{CaptureHandleType/origin}} or
+            {{CaptureHandleType/handle}}.
+            (However, applications should be aware that once a value is read by
+            a capturing application, it may be retained locally by that
+            application.)</p>
           </dd>
         </dl>
       </section>
@@ -810,6 +835,7 @@ dictionary DisplayMediaStreamConstraints {
   DOMString displaySurface;
   boolean logicalSurface;
   DOMString cursor;
+  CaptureHandleType? captureHandle;
 };</pre>
           <dl data-link-for="MediaTrackSettings" data-dfn-for="MediaTrackSettings" class=
           "dictionary-members">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -367,6 +368,29 @@ <h2>
...
+            tracks that would result if the tab in which the application is 
+            running is captured.<p>
+            <p>Only the top-level document may set or read this value. Attempts
+            to read or write the value from other contexts should result in
+            a {{NotAllowedError}}.
+            </p>
+            <p>If the value is empty (unset, set <code>undefined</code> or set 
+            to the empty string), the application leaves the choice of a
+            label up to the user agent.</p>
+            <p>If set to any non-empty value, the user agent SHOULD incorporate
```
guidou: UA should also be able to sanitize it in other ways (e..g, if the label is too long)

index.html
```diff
@@ -367,6 +368,29 @@ <h2>
...
+        <dl data-link-for="MediaDevices" data-dfn-for="MediaDevices">
+          <dt>
+            <dfn>preferredLabel</dfn>
+          </dt>
+          <dd>
+            <p>This settable attribute allows an application to inform the user
+            agent of a meaningful {{MediaStreamTrack/label}} to apply to the
+            tracks that would result if the tab in which the application is 
+            running is captured.<p>
+            <p>Only the top-level document may set or read this value. Attempts
```
guidou: Do you know if there are other ways for a doc to know if it's a top-level doc?
Anyway, I think saying that only the preferredLabel of the top-level document can be used as label would be enough. WDYT?

eladalon1983: I think `window === window.top` is an easier way for a document to check whether it's the top-level document, so we'd not be exposing that bit of information for the first time.

Your suggestion makes sense to me; I will enact the change.

eladalon1983:
This PR is out of date. I will produce a standalone document that more accurately reflects the direction this ended up taking. In the meantime, please see [this explainer](https://docs.google.com/document/d/1oSDmBPYVlxFJxb7ZB_rV6yaAaYIBFDphbkx5bXLnzFg) for what is currently in origin trial in Chrome.

PR #164, opened 2021-04-08
Define the suppressLocalAudioPlayback constraint

### LINKED ISSUE #161: Conditional Suppression of Local Audio Playback from Captured Tab
Consider a user with two open tabs - one with a VC app and one displaying some media. (E.g. Meet+YouTube, Teams+Twitter, or Zoom+Facebook.) Assume the VC tab is capturing the media-playing-tab using `getDisplayMedia` and sharing the resulting audio and video with remote participants. Sometimes it's desirable to suppress the captured tab's audio-playback over the local speakers, without preventing the audio from being captured. That is, the audio is is still captured, but is no longer routed to the local speakers. (Note, btw, that Chrome's hangouts-extension-based API does this very thing.) Rationale:
1. The user could be driving audio from the captured tab to a separate set of speakers in the room in which they're sitting together with a few other people. Meet, for example, has a [present-only mode](https://meet.google.com/?present=1), where the user joins a meeting for presenting only, and does not receive audio. In such a case, the user would not normally expect they have to mute their local machine's speakers via the OS.
2. Meet has found that echo cancellation works better if the media-tab's captured audio is mirrored back to the user and is played back in the VC tab as though it were another participant's audio. Other VC apps likely find themselves in a similar situation.

I propose adding a way for an app to conditionally request that the captured tab's audio playback be suppressed while the capture is ongoing. We could introduce a new constraint called `suppressLocalAudioPlayback` and say that the user agent SHOULD/MAY respect this request. (In the case of multiple capturers - suppress local playback as long as at least one active capture exists with this particular constraint.)

eladalon1983:
Fixes #161. Define suppressLocalAudioPlayback.

***
<a href="https://pr-preview.s3.amazonaws.com/eladalon1983/mediacapture-screen-share/pull/164.html" title="Last updated on Jun 3, 2021, 11:26 AM UTC (624980f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/164/e2a35b3...eladalon1983:624980f.html" title="Last updated on Jun 3, 2021, 11:26 AM UTC (624980f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -570,6 +570,33 @@
                 again.</p>
               </td>
             </tr>
+            <tr id="def-constraint-suppressLocalAudioPlayback">
+              <td><dfn>suppressLocalAudioPlayback</dfn></td>
+              <td>{{ConstrainBoolean}}</td>
+              <td>
+                <p>As a setting, this value indicates whether or not the user
+                agent is applying <a>local audio playback suppression</a> to
+                the source.</p>
+                <p>As a constraint, this value is only meaningful if the user
+                selects capturing a [=browser=] [=display surface=]. In that
+                case, a value of <code>true</code> indicates that the user
+                agent SHOULD perform <a>local audio playback suppression</a>
+                on the captured [=browser=] [=display surface=].</p>
+                <p>When <dfn>local audio playback suppression</dfn> is applied,
+                the user agent SHOULD stop relaying audio to the local speakers,
+                but that audio MUST still be captured by any ongoing audio-capturing
+                capture-sessions. This suppression MUST NOT be observable to the
+                captured document. Furthermore, the capturing document may only
+                observe whether it is applying <a>suppressLocalAudioPlayback</a>;
+                not whether that suppression is having an effect (i.e. can't
+                observe if the user is overriding this in the user agent).</p>
+                <p>When a [=browser=] [=display surface=] is subject to multiple
+                concurrent captures, <a>local audio playback suppression</a>
+                SHOULD be applied as long as at least one active audio-capturing
+                capture-session is constraining <a>suppressLocalAudioPlayback</a>
+                to <code>true</code>.</p>
+              </td>
+            </tr>
           </tbody>
         </table>  
         <p>When inherent properties of the underlying source of a user-selected
@@ -702,6 +729,7 @@ dictionary DisplayMediaStreamConstraints {
   boolean logicalSurface = true;
   boolean cursor = true;
   boolean restrictOwnAudio = true;
+  boolean suppressLocalAudioPlayback = true;
 };</pre>
           <dl  data-dfn-for="MediaTrackSupportedConstraints"
           class="dictionary-members">
@@ -738,6 +766,14 @@ dictionary DisplayMediaStreamConstraints {
                 recognized.
               </p>
             </dd>
+            <dt>
+              <dfn>suppressLocalAudioPlayback</dfn> of type {{boolean}}, defaulting to <code>true</code>
+            </dt>
+            <dd>
+              <p>
+                Whether {{suppressLocalAudioPlayback}} constraint is recognized.
+              </p>
+            </dd>
           </dl>
         </section>
         <section>
@@ -753,6 +789,7 @@ dictionary DisplayMediaStreamConstraints {
   ConstrainBoolean logicalSurface;
   ConstrainDOMString cursor;
   ConstrainBoolean restrictOwnAudio;
+  ConstrainBoolean suppressLocalAudioPlayback;
 };</pre>
           <dl data-dfn-for="MediaTrackConstraintSet" class=
           "dictionary-members">
@@ -791,6 +828,15 @@ dictionary DisplayMediaStreamConstraints {
               <p>
                 This constraint is only applicable to audio tracks. See
                 {{restrictOwnAudio}}.
+              </p>
+            </dd>
+            <dt>
+              <dfn>suppressLocalAudioPlayback</dfn> of type {{ConstrainBoolean}}
+            </dt>
+            <dd>
+              <p>
+                This constraint is only applicable to audio tracks. See
+                {{suppressLocalAudioPlayback}}.
               </p>
             </dd>
           </dl>


index.html
```diff
@@ -570,6 +570,25 @@ <h2 id="constrainable-properties">
...
+              <td>{{ConstrainBoolean}}</td>
+              <td>
+                <p>As a constraint, this value is only meaningful if the user
+                selects capturing audio+video from a tab.
+                In that case, a value of <code>true</code> indicates that the
+                user agent SHOULD perform <a>local audio playback
+                suppression</a> on the captured tab.</p>
+
+                <p><dfn>Local audio playback suppression</dfn> means that
+                the user agent stops relaying audio to the local speakers,
```
jan-ivar: We should be normative here (MUST or SHOULD). I vote for SHOULD, based on https://github.com/w3c/mediacapture-screen-share/issues/161#issuecomment-812764675. I'm also happy to clarify that use case in a note if necessary.

index.html
```diff
@@ -570,6 +570,25 @@ <h2 id="constrainable-properties">
...
             </tr>
+            <tr id="def-constraint-suppressLocalAudioPlayback">
+              <td><dfn>suppressLocalAudioPlayback</dfn></td>
+              <td>{{ConstrainBoolean}}</td>
+              <td>
+                <p>As a constraint, this value is only meaningful if the user
+                selects capturing audio+video from a tab.
+                In that case, a value of <code>true</code> indicates that the
+                user agent SHOULD perform <a>local audio playback
+                suppression</a> on the captured tab.</p>
```
jan-ivar: I'd prefer we mimic the [restrictOwnAudio](https://w3c.github.io/mediacapture-screen-share/#dfn-restrictownaudio) prose with its _"As a setting,"_  and _"As a constraint,"_ paragraphs almost verbatim, and move further explanation of limitations to the linked algorithm. With one potential exception: we might want the dictionary member to be absent on tracks that don't support it. I wasn't able to confirm how `restrictOwnAudio` works in Chrome here, since it doesn't appear to support audio capture outside of tabs, but we should probably have that work the same way, in case a user agents someday return an audio track capturing a non-browser display-surface.

Also, s/tab/[browser display-surface](https://w3c.github.io/mediacapture-screen-share/#dfn-browser)/

We don't need to say "audio+video" since this is the audio constraints table (and audio cannot be captured without video).

I'm unsure about the "local" term, since its opposite is remote, with both having overloaded connotations in most screen-sharing use cases. But I'm also having trouble thinking of better names. What do you think about s/local/target/ ?

"Speakers" may work, hopefully it's clear this applies to non-default audio output devices set by `setSinkId` as well.

Lastly, I'd like a paragraph stating this MUST NOT be observable by the document(s) in the captured browser display-surface. This should clarify this is a UA mute type of thing — e.g. wouldn't affect a site rendering an audio visualizer.

index.html
```diff
@@ -570,6 +570,25 @@ <h2 id="constrainable-properties">
...
+
+                <p><dfn>Local audio playback suppression</dfn> means that
+                the user agent stops relaying audio to the local speakers,
+                but that audio is still captured by any ongoing audio-capturing
+                sessions. In the case of multiple concurrent captures of the
+                same tab, local audio playback suppression will be active if and
+                only if at least one of the active sessions specified
```
jan-ivar: I'd replace "if and only if" phrasing with SHOULD here as well.

jan-ivar (changes requested):
Overall LGTM. Some nits and questions around behavior.

Note that constraints tables are perhaps a bit unusual in that they're normative, serving as the sole description of desired behavior outside of the general constraints algorithm in mediacapture-main. This changes the audience a bit.

index.html
jan-ivar: Let's keep the name `suppressLocalAudioPlayback` though, since `suppressTargetAudioPlayback` might be misunderstood to mean muting in the capture. Also, the "Local" overloaded connotation somewhat fits in the most common use case.

We could still use "target" in implementer prose, e.g. rename the algorithm to <a>target audio playback suppression</a>, but I don't have a strong opinion there either.

eladalon1983:
> Overall LGTM. Some nits and questions around behavior.
> 
> Note that constraints tables are perhaps a bit unusual in that they're normative, serving as the sole description of desired behavior outside of the general constraints algorithm in mediacapture-main. This changes the audience a bit.

Thanks for reviewing. IIUC, you'd still like this to be presented and discussed in the upcoming WebRTC WG interim meeting, so I will hold off on affecting these changes until we've discussed it there.

index.html
eladalon1983: IMHO it would be better to use MUST, but we can start with SHOULD and consider updating in the future, I think.

index.html
eladalon1983: > I'm unsure about the "local" term, since its opposite is remote, with both having overloaded connotations in most screen-sharing use cases. But I'm also having trouble thinking of better names. What do you think about s/local/target/ ?

I think "local" is the lesser evil, especially as you agree we should keep the name `suppressLocalAudioPlayback`. I don't think it would be reasonable to use a different term in the description and expect readers to understand that no distinction between "local" and "target" is intended. One concept - one word; at least within this section.

> Lastly, I'd like a paragraph stating this MUST NOT be observable by the document(s) in the captured browser display-surface. This should clarify this is a UA mute type of thing — e.g. wouldn't affect a site rendering an audio visualizer.

Is this meant as future-proofing, or is there some existing mechanism for observing this state?

index.html
eladalon1983: Thanks for reviewing. I've updated the PR - PTAL.

eladalon1983:
Thanks for reviewing. I've updated the PR - PTAL.

index.html
```diff
@@ -570,6 +570,32 @@ <h2 id="constrainable-properties">
...
+                agent is applying <a>local audio playback suppression</a> to
+                the source.</p>
+                <p>As a constraint, this value is only meaningful if the user
+                selects capturing a [=browser=] [=display surface=]. In that
+                case, a value of <code>true</code> indicates that the user
+                agent SHOULD perform <a>local audio playback suppression</a>
+                on the captured [=browser=] [=display surface=].</p>
+                <p>When <dfn>local audio playback suppression</dfn> is applied,
+                the user agent SHOULD stop relaying audio to the local speakers,
+                but that audio MUST still captured by any ongoing audio-capturing
```
```suggestion
                but that audio MUST still be captured by any ongoing audio-capturing
```

index.html
```diff
@@ -570,6 +570,32 @@ <h2 id="constrainable-properties">
...
+                capture-sessions.</p>
+                <p>When a [=browser=] [=display surface=] is subject to multiple
+                concurrent captures, <a>local audio playback suppression</a>
+                SHOULD be active as long as at least one active audio-capturing
```
```suggestion
                SHOULD be applied as long as at least one active audio-capturing
```

index.html
```diff
@@ -570,6 +570,32 @@ <h2 id="constrainable-properties">
...
+                capture-session remains which specified
+                <a>suppressLocalAudioPlayback</a>.
```
```suggestion
                capture-session is constraining <a>suppressLocalAudioPlayback</a>
                to <code>true</code>.
```

index.html
```diff
@@ -570,6 +570,32 @@ <h2 id="constrainable-properties">
...
+                (Note that capture-sessions for which the user did not permit
+                audio-capture are not considered audio-capturing for this
+                purpose.)</p>
```
jan-ivar: I think the word "active" in "active audio-capturing" above covers this, so we don't need this, I don't think.

index.html
```diff
@@ -570,6 +570,32 @@ <h2 id="constrainable-properties">
...
+                the source.</p>
+                <p>As a constraint, this value is only meaningful if the user
+                selects capturing a [=browser=] [=display surface=]. In that
+                case, a value of <code>true</code> indicates that the user
+                agent SHOULD perform <a>local audio playback suppression</a>
+                on the captured [=browser=] [=display surface=].</p>
+                <p>When <dfn>local audio playback suppression</dfn> is applied,
+                the user agent SHOULD stop relaying audio to the local speakers,
+                but that audio MUST still captured by any ongoing audio-capturing
+                capture-sessions.</p>
```
jan-ivar: I'd still like the following sentence in about observability in JavaScript, to clarify what should be implemented, since I see two interpretations otherwise:
 1. Speaker output is muted at the tab-level in the browser, with no behavioral impact to the JS documents in it.
 2. Audio observably stops in APIs like tracks and web audio.  

I want to clarify we mean #1. E.g. any graphical audio visualizer a page makes up, keeps animating, and any audio it sends over a peer connection will still be audible at the peer.

```suggestion
                capture-sessions. This suppression MUST not be observable to the document,
                other than through its setting value.</p>
```

jan-ivar (changes requested):
Thanks, I'd still like a sentence about observability in JavaScript.

index.html
eladalon1983: Reworded this somewhat.
1. MUST NOT
2. Clarified observability to the captured document (none at all) vs. to the capturing document (can only re-examine if it has applied the constraint; not if it has/had an effect).

eladalon1983:
Changes made. @jan-ivar, PTAL.

jan-ivar (approved):

index.html
```diff
@@ -570,6 +570,33 @@ <h2 id="constrainable-properties">
...
+              <td>{{ConstrainBoolean}}</td>
+              <td>
+                <p>As a setting, this value indicates whether or not the user
+                agent is applying <a>local audio playback suppression</a> to
+                the source.</p>
+                <p>As a constraint, this value is only meaningful if the user
+                selects capturing a [=browser=] [=display surface=]. In that
+                case, a value of <code>true</code> indicates that the user
+                agent SHOULD perform <a>local audio playback suppression</a>
+                on the captured [=browser=] [=display surface=].</p>
```
mickb47lol, line 584: []()

mickb47lol (commented):
- [ ]

PR #170, opened 2021-05-27
Remove restriction over changing capture source

### LINKED ISSUE #81: The user agent should be allowed to change sources after getDisplayMedia() resolves
The spec currently has language saying the source is constant after getDisplayMedia() returns a stream, e.g:

> The provided media MUST include precisely one track of each media type in requestedMediaTypes. The devices chosen MUST be the ones determined by the user. Once selected, the source of a MediaStreamTrack MUST NOT change.
> [...]
> Since the source of media cannot be changed after a MediaStreamTrack has been returned and constraints do not affect the selection of display surfaces, these constrainable properties cannot be changed by an application.

This refers to the settings "displaySurface", "logicalSurface" or "cursor" to be constant for a track. I'm not convinced we should enforce these being constants, but even if we do want that we don't have to go so far to say sources are constant, just that sources cannot change type from "logical" to "display" or enable/disable cursor. And just because a user agent (end-user) can change the source doesn't mean that the application has any control of this.

We should allow the user agent to change source on-the-fly. This allows, amongst other things, the user agent to allow the end-user to change which tab ("browser" surface) to present at any given point in time. This is what current extensions are already doing (edit: actually I'm not sure if this is already implemented or just planned to be implemented).

eladalon1983:
Addresses issue #81.

***
<a href="https://pr-preview.s3.amazonaws.com/eladalon1983/mediacapture-screen-share/pull/170.html" title="Last updated on May 27, 2021, 2:59 PM UTC (dd6b808)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/170/556ec65...eladalon1983:dd6b808.html" title="Last updated on May 27, 2021, 2:59 PM UTC (dd6b808)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -299,7 +299,8 @@
                     each media type in <var>requestedMediaTypes</var>.
                     The devices chosen MUST be the ones determined by the user.
                     Once selected, the source of a
-                    {{MediaStreamTrack}} MUST NOT change.</p>
+                    {{MediaStreamTrack}} MUST NOT change, unless the user
+                    permits it through their interaction with the user agent.</p>
                     <p>User Agents are encouraged to warn users against sharing
                     <a>browser</a> display devices as well as <a>monitor</a>
                     display devices where browser windows are visible, or


index.html
```diff
@@ -297,9 +297,7 @@ <h2>
                     provided media.</p>
                     <p>The provided media MUST include precisely one track of
                     each media type in <var>requestedMediaTypes</var>.
-                    The devices chosen MUST be the ones determined by the user.
-                    Once selected, the source of a
-                    {{MediaStreamTrack}} MUST NOT change.</p>
+                    The devices chosen MUST be the ones determined by the user.</p>
```
jan-ivar: Can we say something like MUST NOT CHANGE except with the user's permission.

eladalon1983: We can. PTAL at the new text - wdyt?

jan-ivar (approved):

Salam01718857:
Please accept me.

eladalon1983:
> Please accept me.

You are beautiful just as you are.

PR #183, opened 2021-06-04
chore: use specprod v2

marcoscaceres:


dontcallmedom (changes requested):
moving away from the manifest-based approach requires adding the options for W3C_BUILD_OVERRIDE

marcoscaceres:
Ah, never mind then :)

PR #206, opened 2022-01-24
Add note and example of inaccessible stats that fire mute

### LINKED ISSUE #141: Should mute and unmute events of MediaStreamTrack be allowed to fire based on user non-action?
The specification uses the term "user activation" exactly once

> When the getDisplayMedia() method is called, the User Agent MUST run the following steps:
> 
> If the method call is not triggered by user activation, return a promise rejected with a DOMException object whose name attribute has the value InvalidStateError.

The terms "user action" and "user gesture" are not included in the language of the specification.

Chromium 85 appears to fire `mute` and `unmute` events on `MediaStreamTrack` from `getDisplayMedia()` directly corresponding to user non-action, for example, not moving the cursor on the captured screen, or user action, moving the cursor on the captured screen.

```
      onclick = async e => {
        var input,
          recorder,
          audioTrack,
          videoTrack,
          stream,
          mediaStream,
          chunks = [];

        onclick = null;

        stream = await navigator.mediaDevices.getDisplayMedia({ video: true });

        [videoTrack] = stream.getVideoTracks();
        videoTrack.onended = videoTrack.onmuted = e => console.log(e);
        await videoTrack.applyConstraints({
          resizeMode: 'none',
          cursor: 'never',
          width: window.innerWidth * 0.7,
          height: window.innerHeight,
        });
        video = document.createElement('video');
        video.currentTime = 0;
        video.autoplay = true;
        videoTrack.onunmute = videoTrack.onmute = e => console.log(e);

        video.ontimeupdate = _ => {
          console.log(video.currentTime, 60 * 7 + 60 * 0.5);
          if (video.currentTime > 60 * 7 + 60 * 0.5) {
            recorder.stop();
            videoTrack.stop();
          }
        };
        video.onplay = _ => {
          recorder = new MediaRecorder(stream);
          recorder.start(0);
          recorder.onstop = async e => {
            console.log(
              URL.createObjectURL(new Blob(chunks, { type: 'video/webm' }))
            );
          };
          recorder.ondataavailable = e => {
            if (e.data.size > 0) chunks.push(e.data);
          };
        };
        video.srcObject = stream;
      };
```

The result is a series of unintended consequences impacting other API's, including media file produced by `MediaRecorder`, `timeupdate` event of `HTMLMediaElement` not firing every 50 to 250ms per HTML Standard. The resulting bugs that have  observed so far downstream https://bugs.chromium.org/p/chromium/issues/detail?id=1099280.

Kindly include language in the specification which prohibits implementations from firing `mute` and `unmute` events based on user non-action or user action.

jan-ivar:
Fixes https://github.com/w3c/mediacapture-screen-share/issues/141.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/mediacapture-screen-share/pull/206.html" title="Last updated on Feb 3, 2022, 3:49 PM UTC (a3aa5ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/206/f725538...jan-ivar:a3aa5ae.html" title="Last updated on Feb 3, 2022, 3:49 PM UTC (a3aa5ae)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -379,8 +379,13 @@
           A <a>display surface</a> that is being shared may temporarily or permanently become
           inaccessible to the application because of actions taken by the operating system or user
           agent. What makes a <a>display surface</a> considered inaccesible is outside the scope
-          of this specification, but examples MAY include a <a>monitor</a> disconnecting or an
-          <a>application</a>, <a>window</a> or <a>browser</a> closing or becoming minimized.
+          of this specification, but examples MAY include a <a>monitor</a> disconnecting, an
+          <a>application</a>, <a>window</a> or <a>browser</a> closing or becoming minimized,
+          or due to an incoming call on a phone.
+        </p>
+        <p class="note">
+          User agents ultimately control what inaccesible means in this context, but are expected
+          to only fire mute and unmute events for interruptions that have external reasons.
         </p>
         <p>
           When <a>display surface</a> enters an inaccessible state that is not necessarily
--- a/index.html
+++ b/index.html
@@ -384,7 +384,7 @@
           or due to an incoming call on a phone.
         </p>
         <p class="note">
-          User agents ultimately control what inaccesible means in this context, but are expected
+          User agents ultimately control what inaccesible means in this context, but are encouraged
           to only fire mute and unmute events for interruptions that have external reasons.
         </p>
         <p>


index.html
```diff
@@ -379,8 +379,13 @@ <h2 id="hidden-display-surfaces">
...
           agent. What makes a <a>display surface</a> considered inaccesible is outside the scope
-          of this specification, but examples MAY include a <a>monitor</a> disconnecting or an
-          <a>application</a>, <a>window</a> or <a>browser</a> closing or becoming minimized.
+          of this specification, but examples MAY include a <a>monitor</a> disconnecting, an
+          <a>application</a>, <a>window</a> or <a>browser</a> closing or becoming minimized,
+          or due to an incoming call on a phone.
+        </p>
+        <p class="note">
+          User agents ultimately control what inaccesible means in this context, but are expected
+          to only fire mute and unmute events for interruptions that have external reasons.
```
youennf, line 388: I do not really understand what 'interruptions that have external reasons' mean so I am not sure how this helps.
The wording does seem to exclude valid cases like:
- A user interacting with User Agent UI to temporarily mute capture.
- A user agent muting a track because another page started to capture a screen.
- A user agent muting a track after a large timer, like 24 hours.

eladalon1983: I think a good separation would be:
* The **capturing document**, by which I mean the entity holding the track, can directly pause the capture via [enabled](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/enabled).
* Every other cause else should be reflected through [muted](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/muted). Regardless of it is directly/indirectly caused by the capturing document.

For distinguishing different mute-causes I [have suggested](https://github.com/w3c/mediacapture-region/issues/9#issuecomment-1022334881) this:
```webidl
enum MuteCasue {
  "user",
  "target_lost",
  "zero_pixel_frame",
  ...
};

partial interface MediaStreamTrack {
  readonly sequence<MuteCasue> mute_causes;
}
```

index.html
```diff
@@ -379,8 +379,13 @@ <h2 id="hidden-display-surfaces">
           A <a>display surface</a> that is being shared may temporarily or permanently become
           inaccessible to the application because of actions taken by the operating system or user
           agent. What makes a <a>display surface</a> considered inaccesible is outside the scope
-          of this specification, but examples MAY include a <a>monitor</a> disconnecting or an
-          <a>application</a>, <a>window</a> or <a>browser</a> closing or becoming minimized.
+          of this specification, but examples MAY include a <a>monitor</a> disconnecting, an
+          <a>application</a>, <a>window</a> or <a>browser</a> closing or becoming minimized,
```
eladalon1983, line 383: Drive-by: As far as I can tell (from [this test](https://jsfiddle.net/k8wLn7uv/17/)), Firefox does not mute minimized windows (at least on Mac). Safari does not yet support windows/tabs, so I am unsure what their intentions there are. As long as we don't actually have a UA that either does so or intends to do so, can we get rid of this example?

Drive-by^2: Btw, @jan-ivar, I see the odd pattern below when minimizing a window that's being captured. It seems like the minimization animation on Mac is being captured by Firefox, then gets stuck on the last frame. Any idea if this is a known issue?
Screen Shot 2022-01-31 at 14 24 48

index.html
```diff
@@ -379,8 +379,13 @@ <h2 id="hidden-display-surfaces">
...
+          or due to an incoming call on a phone.
+        </p>
+        <p class="note">
+          User agents ultimately control what inaccesible means in this context, but are expected
```
```suggestion
          User agents ultimately control what inaccesible means in this context, but are encouraged
```

index.html
jan-ivar: Sounds like a new issue, but not super excited fwiw

PR #216, opened 2022-03-22
Define DisplayMediaStreamConstraints.selfBrowserSurface

### LINKED ISSUE #209: Avoid Hall-of-Mirrors
Screen Shot 2022-02-22 at 20 23 23

**Accidental** self-capture is a common problem for video conferencing software. When users accidentally choose the tab in which the VC app is running, a Hall-of-Mirrors effect is produced, confusing users and derailing discussions with remote users. Eliminating this would improve the lives of users and Web-developers alike.

When `getDisplayMedia()` is called, the user agent is supposed to show an unconstrained list of all possible surfaces, and that includes the tab from which `getDisplayMedia` was called. However, members of this Working Group have in the past expressed a desire to exclude the current tab from this list, with the rationale that it's an attack-vector for bypassing origin separation. (That is to say - an application capturing itself can embed arbitrary content and see it.) I intend to propose a mechanism in this vein. The mechanism I am about to propose does **not** purport to address the security concern. But I argue that it offers **no degradation in security**.

The first option to consider is that browsers could simply eliminate the option without a change in spec. However, this solution is unlikely to be universally adopted, as some legitimate applications currently use self-capture, and some existing browser would be apprehensive of breaking them (e.g. Chrome).

We could (and should) step in and help bridge the gap, for the betterment of the Web platform. I propose the following:

Recall that `getDisplayMedia`'s argument is of type [DisplayMediaStreamConstraints](https://w3c.github.io/mediacapture-screen-share/#displaymediastreamconstraints). That is:
```webidl
partial interface MediaDevices {
  Promise<MediaStream> getDisplayMedia(optional DisplayMediaStreamConstraints constraints = {});
};
```

I propose that we extend [DisplayMediaStreamConstraints](https://w3c.github.io/mediacapture-screen-share/#displaymediastreamconstraints) with `excludeCurrentTab`:
```webidl
dictionary DisplayMediaStreamConstraints {
  (boolean or MediaTrackConstraints) video = true;
  (boolean or MediaTrackConstraints) audio = false;
  boolean excludeCurrentTab = ?;  // Default value - let's discuss.
};
```

If `excludeCurrentTab` is set to `true`, the user agent MAY/SHOULD/MUST exclude the current tab from the list of tabs.
If `excludeCurrentTab` is set to `false`, the user agent does as it wishes. (Chrome will include the current tab, as it does today, but other browsers are free to exclude the current tab either way.)

eladalon1983:
Fixes #209.

***
<a href="https://pr-preview.s3.amazonaws.com/eladalon1983/mediacapture-screen-share/pull/216.html" title="Last updated on Apr 7, 2022, 3:58 PM UTC (7aa21ce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/216/edc7772...eladalon1983:7aa21ce.html" title="Last updated on Apr 7, 2022, 3:58 PM UTC (7aa21ce)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -722,6 +722,62 @@
           </div>
         </section>
         <section>
+          <h2><dfn>CurrentTabPreferenceEnum</dfn></h2>
+          <p>
+            Recall that when offering the user a choice of [=display surfaces=]
+            to capture, the user agent MAY offer the [=browser=] [=display surface=]
+            in which the application is running. An application might prefer that
+            this particular [=display surface=] not be offered to the user.
+            The CurrentTabPreferenceEnum enumeration describes the different
+            preferences an application might have.
+          </p>
+          <div class="note">
+            <p>
+              For clarity's sake, "tab" is used rather than "browser display surface".
+            </p>
+          </div>
+          <pre class="idl">
+            enum CurrentTabPreferenceEnum {
+              "none",
+              "include",
+              "exclude"
+            };
+          </pre>
+          <table  data-dfn-for="CurrentTabPreferenceEnum" class="simple">
+            <tbody>
+              <tr>
+                <th colspan="2">
+                  Enumeration description
+                </th>
+              </tr>
+              <tr>
+                <td>
+                  <dfn id="idl-def-CurrentTabPreferenceEnum.none">none</dfn>
+                </td>
+                <td>
+                  No preference by the application.
+                </td>
+              </tr>
+              <tr>
+                <td>
+                  <dfn id="idl-def-CurrentTabPreferenceEnum.include">include</dfn>
+                </td>
+                <td>
+                  An application preference of including the current tab.
+                </td>
+              </tr>
+              <tr>
+                <td>
+                  <dfn id="idl-def-CurrentTabPreferenceEnum.exclude">exclude</dfn>
+                </td>
+                <td>
+                  An application preference of excluding the current tab.
+                </td>
+              </tr>
+            </tbody>
+          </table>
+        </section>
+        <section>
           <h2>DisplayMediaStreamConstraints</h2>
           <p>The <dfn>DisplayMediaStreamConstraints</dfn> dictionary is used to
           instruct the user agent what sort of {{MediaStreamTrack}}s may be
@@ -732,6 +788,7 @@
 dictionary DisplayMediaStreamConstraints {
   (boolean or MediaTrackConstraints) video = true;
   (boolean or MediaTrackConstraints) audio = false;
+  CurrentTabPreferenceEnum currentTabPreference = "none";
 };</pre>
             <section>
               <h2>Dictionary <a class="idlType">DisplayMediaStreamConstraints</a>
@@ -760,6 +817,16 @@ dictionary DisplayMediaStreamConstraints {
                   further specifies desired processing options to be applied to
                   the audio track. If <code>false</code>, the {{MediaStream}}
                   will not contain an audio track.</p>
+                </dd>
+                <dt>
+                  <dfn><code>currentTabPreference</code></dfn> of type {{CurrentTabPreferenceEnum}},
+                  defaulting to {{CurrentTabPreferenceEnum/none}}
+                </dt>
+                <dd>
+                  Signals application preference for whether the current
+                  [=browser=] [=display surface=] should be included in the list
+                  of [=display surfaces=] offered to the user.
+                  The user agent MAY regard this hint.
                 </dd>
               </dl>
             </section>
--- a/index.html
+++ b/index.html
@@ -759,7 +759,8 @@
                   <dfn id="idl-def-CurrentTabPreferenceEnum.include">include</dfn>
                 </td>
                 <td>
-                  An application preference of including the current tab.
+                  The application prefers the tab it is in to be included
+                  in the choices offered.
                 </td>
               </tr>
               <tr>
--- a/index.html
+++ b/index.html
@@ -808,9 +808,9 @@ dictionary DisplayMediaStreamConstraints {
                 </dt>
                 <dd>
                   If present, signals application preference for whether the
-                  [=browser=] [=display surface=] associated with
+                  [=browser=] [=display surface=] which is associated with
                   [=this=]'s [=relevant global object=]'s [=associated Document=]'s
-                  [=top-level browsing context=] should be among the choices
+                  [=top-level browsing context=], should be among the choices
                   offered to the user. The user agent MAY ignore this hint.
                 </dd>
               </dl>
--- a/index.html
+++ b/index.html
@@ -768,7 +768,8 @@
                   <dfn id="idl-def-CurrentTabPreferenceEnum.exclude">exclude</dfn>
                 </td>
                 <td>
-                  An application preference of excluding the current tab.
+                  The application prefers the tab it is in to be excluded
+                  from the choices offered.
                 </td>
               </tr>
             </tbody>
--- a/index.html
+++ b/index.html
@@ -773,7 +773,7 @@
 dictionary DisplayMediaStreamConstraints {
   (boolean or MediaTrackConstraints) video = true;
   (boolean or MediaTrackConstraints) audio = false;
-  SelfCapturePreferenceEnum selfBrowserSurface = "none";
+  SelfCapturePreferenceEnum selfBrowserSurface;
 };</pre>
             <section>
               <h2>Dictionary <a class="idlType">DisplayMediaStreamConstraints</a>
@@ -804,8 +804,7 @@ dictionary DisplayMediaStreamConstraints {
                   will not contain an audio track.</p>
                 </dd>
                 <dt>
-                  <dfn><code>selfBrowserSurface</code></dfn> of type {{SelfCapturePreferenceEnum}},
-                  defaulting to {{SelfCapturePreferenceEnum/none}}
+                  <dfn><code>selfBrowserSurface</code></dfn> of type {{SelfCapturePreferenceEnum}}
                 </dt>
                 <dd>
                   If present, signals application preference for whether the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -761,6 +818,16 @@ <h2>Dictionary <a class="idlType">DisplayMediaStreamConstraints</a>
...
                 </dd>
+                <dt>
+                  <dfn><code>currentTabPreference</code></dfn> of type {{CurrentTabPreferenceEnum}},
+                  defaulting to {{CurrentTabPreferenceEnum/none}}
+                </dt>
+                <dd>
+                  Signals application preference for whether the current
+                  [=browser=] [=display surface=] should be included in the list
+                  of [=display surfaces=] offered to the user.
+                  The user agent MAY regard this hint.
```
jan-ivar: "Current" is not defined. Also reverse the MAY to clarify this prose is optional:
```suggestion
                  If present, signals application preference for whether the
                  [=browser=] [=display surface=] associated with
                  [=this=]'s [=relevant global object=]'s [=associated Document=]'s
                  [=top-level browsing context=] should be among the choices
                  offered to the user. The user agent MAY ignore this hint.
```

index.html
```diff
@@ -721,6 +721,62 @@ <h2>
...
         </section>
+        <section>
+          <h2><dfn>CurrentTabPreferenceEnum</dfn></h2>
+          <p>
+            Recall that when offering the user a choice of [=display surfaces=]
+            to capture, the user agent MAY offer the [=browser=] [=display surface=]
+            in which the application is running. An application might prefer that
+            this particular [=display surface=] not be offered to the user.
+            The CurrentTabPreferenceEnum enumeration describes the different
+            preferences an application might have.
```
jan-ivar: I'd shorten this to not repeat normative language below. How about:
```suggestion
            Describes the different hints an application can provide about whether the
            tab the application is in should be among the choices offered to the user.
```

index.html
```diff
@@ -721,6 +721,62 @@ <h2>
...
+          </p>
+          <div class="note">
+            <p>
+              For clarity's sake, "tab" is used rather than "browser display surface".
+            </p>
+          </div>
```
jan-ivar: I don't think we need this note, since this prose is largely explanatory. I think it's fine to use "tab" here as a shorthand, as long as normative prose below doesn't use it and sticks to defined terms only.

index.html
```diff
@@ -721,6 +721,62 @@ <h2>
...
+                <td>
+                  No preference by the application.
+                </td>
+              </tr>
+              <tr>
+                <td>
+                  <dfn id="idl-def-CurrentTabPreferenceEnum.include">include</dfn>
+                </td>
+                <td>
+                  An application preference of including the current tab.
```
```suggestion
                  The application prefers the tab it is in to be included
                  in the choices offered.
```

index.html
```diff
@@ -721,6 +721,62 @@ <h2>
...
+                </td>
+              </tr>
+              <tr>
+                <td>
+                  <dfn id="idl-def-CurrentTabPreferenceEnum.exclude">exclude</dfn>
+                </td>
+                <td>
+                  An application preference of excluding the current tab.
```
```suggestion
                  The application prefers the tab it is in to be excluded
                  from the choices offered.
```

index.html
```diff
@@ -721,6 +721,62 @@ <h2>
...
+            preferences an application might have.
+          </p>
+          <div class="note">
+            <p>
+              For clarity's sake, "tab" is used rather than "browser display surface".
+            </p>
+          </div>
+          <pre class="idl">
+            enum CurrentTabPreferenceEnum {
+              "none",
```
jan-ivar: `currentTabPreference: "none"` might be confused with `currentTabPreference: "exclude"`. `currentTabPreference: "no-preference"` seems clearer.

I can't decide whether this third option is cleaner or redundant, since this is input only. I couldn't find any design guidance on it.

One thing to be aware of is that extending enums later is tricky (because old browsers will throw on new values). So it's easier to add now and remove later than the opposite.
```suggestion
              "no-preference",
```

index.html
```diff
@@ -732,6 +788,7 @@ <h2>DisplayMediaStreamConstraints</h2>
 dictionary DisplayMediaStreamConstraints {
```
jan-ivar, line 773: This new member is not a constraint, so I think this is now officially an options object that among other things houses video and audio constraints: 
```suggestion
dictionary DisplayMediaStreamOptions {
```

index.html
```diff
@@ -732,6 +788,7 @@ <h2>DisplayMediaStreamConstraints</h2>
...
   (boolean or MediaTrackConstraints) video = true;
   (boolean or MediaTrackConstraints) audio = false;
+  CurrentTabPreferenceEnum currentTabPreference = "none";
```
jan-ivar: I could be wrong, but this would be the first time the word "Tab" appears in a web API. I know this was already discussed in the issue, but I can't help thinking we're over-selecting here.

At a conceptual level, the user agent learns whether self-capture (hall of mirrors) is part of the application's use case(s) or not. This seems like valuable information to the user agent, at its most abstract.

It seems less relevant how the user agent ends up acting on this information: whether this applies only to tab-capture or whether it also applies to window capture of the browser window the document is in ... those are really good questions user agents should perhaps be left to figure out, which seems OK to me.

I also don't think it matters if user agents fall short of promises on this today, since this is a hint. E.g. I wouldn't expect browsers to ever exclude full-screen capture over this, but a user agent might take `"exclude"` as a clue it's OK to blur away the undesirable hall-of-mirrors effect for instance. This way the user agent doesn't need to worry it's stepping on apps that rely on self-capture. cc @youennf 

So I'd suggest an iteration on @martinthomson's wording:
```suggestion
  CurrentTabPreferenceEnum selfCapture = "no-preference";
```

jan-ivar (changes requested):

index.html
eladalon1983: [getDisplayMedia](https://w3c.github.io/mediacapture-screen-share/#dom-mediadevices-getdisplaymedia) accepts [DisplayMediaStreamConstraints](https://w3c.github.io/mediacapture-screen-share/#dom-displaymediastreamconstraints), not `DisplayMediaStreamOptions`. It's an unfortunate name, but that is the name. Or are you suggesting that we make [getDisplayMedia](https://w3c.github.io/mediacapture-screen-share/#dom-mediadevices-getdisplaymedia) accept an additional parameter of type `DisplayMediaStreamOptions`? I don't see that in your suggested edits.

index.html
eladalon1983: That's fine by me. We can remove.

index.html
eladalon1983: I'm OK with the s/none/no-preference replacement.
I don't want to accept the s/currentTabPreference/selfCapture suggestion, because this only refers to tabs, and that is intentional - we only got consensus over that in the interim meeting. (Also note how it's CurrentTabPreferenceEnum, btw.)

jan-ivar: (with chair-hat off)

I recall us having rough agreement (not consensus which would require a call on the list) [on](https://www.w3.org/2022/03/15-webrtc-minutes.html#t04) this being a "hint", which seems to leave this more as an indicator describing application preference, than an explicit control surface to enact certain specific behavior.

Yes I'm bikeshedding, and perhaps suggesting a bit of abstraction here that wasn't discussed before, but I don't read the last meeting as a final word on naming, and would love to hear from others. I think the question of whether to introduce the term "Tab" to the web platform seems orthogonal to discussion over behavior.

eladalon1983: > I recall us having rough agreement (not c[on](https://www.w3.org/2022/03/15-webrtc-minutes.html#t04)sensus which would require a call on the list) on this being a "hint", which seems to leave this more as an indicator describing application preference, than an explicit control surface to enact certain specific behavior.

This has not changed. We're discussing what the hint will be called.

> I think the question of whether to introduce the term "Tab" to the web platform seems orthogonal to discussion over behavior.

If you're not convinced that "tab" can be a convenient shorthand in this case, we can go with the following alternative, although I consider it inferior to "tab":
```webidl
  SelfCapturePreference selfCaptureBrowser = "no-preference";

  enum SelfCapturePreference {
    "no-preference",
    "include",
    "exclude",
  };
```

This way:
1. We use the [browser](https://w3c.github.io/mediacapture-screen-share/#dfn-browser) that everyone is already forced to know means "tab" in most relevant contexts.
2. We have not blocked the way for `selfCaptureWindow` to be introduced later, if the WG happens to be interested later.

--
But please see my `note` in the original PR. I think "tab" can be understood well as shorthand for "display surface of type 'browser'". My intention is that this would not formally introduce "tab" into the Web platform, but rather, serve as a localized convenience.

eladalon1983:
Hi @jan-ivar, could you PTAL?

index.html
eladalon1983: Ping @jan-ivar; PTAL.

youennf:
I would rename the property to slefBrowserSurface and remove no-preference.

index.html
```diff
@@ -721,6 +721,55 @@ <h2>
...
+            tab the application is in should be among the choices offered to the user.
+          </p>
+          <pre class="idl">
+            enum CurrentTabPreferenceEnum {
+              "no-preference",
+              "include",
+              "exclude"
+            };
+          </pre>
+          <table  data-dfn-for="CurrentTabPreferenceEnum" class="simple">
```
youennf: two spaces

youennf:
What we plan to do:
- rename the property to selfBrowserSurface
- remove no-preference
- remove the use of tab and use something like 'browser surface' in the PR, including CurrentTabPreferenceEnum.

youennf (approved):
Look good to me with the changes.
I think we should beef up the security section since we are now allowing the web page to fiddle with UA UI.

jan-ivar (approved):

eladalon1983:
Ping @jan-ivar and @youennf with a reminder to double-check the delta from what we agreed on and what I ended up editing and merging.

PR #224, opened 2022-05-16
Add spec-hooks for "actively screen-capturing"

### LINKED ISSUE #213: Spec-hooks requested
Implementers working on other specs have requested (e.g. on [this issue](https://github.com/WICG/compute-pressure/issues/82)) that we add some spec-hooks. For example, some behavior might be allowed/forbidden if a document holds a video track that refers to an active screen-capture. I think this is a sensible addition (keeping out-of-scope **how** these hooks will be used). I propose:
* A document is called an "actively screen-capturing document" if that document holds a reference to a "live" video track derived from a call to getDisplayMedia. (In the future - also getViewportMedia.)
* A user agent is called "actively screen-capturing" if any of its fully-active documents is an actively screen-capturing document.

Thoughts? @alvestrand? @jan-ivar? @youennf?

### LINKED ISSUE #887: Spec-hooks requested

eladalon1983:
Fixes w3c/mediacapture-main#887.

***
<a href="https://pr-preview.s3.amazonaws.com/eladalon1983/mediacapture-screen-share/pull/224.html" title="Last updated on Jun 2, 2022, 3:08 PM UTC (b1a22c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/224/0222d62...eladalon1983:b1a22c3.html" title="Last updated on Jun 2, 2022, 3:08 PM UTC (b1a22c3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -126,6 +126,17 @@
       </p>
 
       <p>The <dfn>devicechange</dfn> event is defined in [[GETUSERMEDIA]].</p>
+
+      <p>
+        A document is called an <dfn class="export">actively screen-capturing document</dfn>
+        if that document holds a reference to a {{MediaStreamTrackState/"live"}} video track
+        derived from a call to {{MediaDevices/getDisplayMedia}}.
+      </p>
+
+      <p>
+        A user agent is called <dfn class="export">actively screen-capturing</dfn> if any of
+        its [=Document/fully active=] documents is an [=actively screen-capturing document=].
+      </p>
     </section>
     <section>
       <h2>


dontcallmedom (approved):

index.html
```diff
@@ -126,6 +126,17 @@ <h2>
       </p>
 
       <p>The <dfn>devicechange</dfn> event is defined in [[GETUSERMEDIA]].</p>
+
+      <p>
+        A document is called an <dfn class="export">actively screen-capturing document</dfn>
+        if that document holds a reference to a {{MediaStreamTrackState/"live"}} video track
+        derived from a call to {{MediaDevices/getDisplayMedia}}.
```
youennf, line 133: Should there be a camera/microphone specific version or should it be a generic version (camera/microphone/display capture)?
The latter would be defined in mediacapture-main by checking if the document  [[devicesLiveMap]] has any value being set to true.

We might also want to take muted into account. If all tracks are muted=true, is the document still actively screen capturing?
enabled=false is also something we should think about.

eladalon1983: @kenchris

kenchris: Sorry missed this. Yes, generic version would be preferred, and I think it makes sense to consider muted

youennf:
Fixed by https://github.com/w3c/mediacapture-main/pull/890

PR #225, opened 2022-05-20
Introduce DisplayMediaStreamConstraints.surfaceSwitching

### LINKED ISSUE #223: Dynamic Screenshare Source (Share-this-surface-instead)
### Problem statement
User who share one display surface often wish to make a **quick** transition to sharing another.

### Example 1
* The user prepares to deliver a presentation over VC.
* Before the call starts, the user sets up N tabs, each pre-navigated to a specific location, relevant application already set up. Note that some setup might be involved here. Maybe the user has set up some of these applications to a very specific state.
* The user begins call over a VC application.
* The user interacts with VC application, then with browser's dialog, to share the first tab.
* The user interacts with the first shared tab.
* The user wishes to switch to second tab.
* **What now?** The user has to reactivate the VC tab, re-engage screensharing, etc. **Very suboptimal!**

### Example 2
Same thing, but now that user is recording a lecture using something like Loom.

### Chrome's current solution
Chrome currently offers an extension API that presents a `Share this tab instead` button.

Screenshot 2022-04-29 at 17 12 45

The user just hops to the next tab, presses the button, and continues the presentation.
0 beats skipped.
This solution is implementable by user agents without any API changes. **In theory.**

### But...
Not all applications are able to smoothly deal with a change of source. What about applications that rely on tight-coupling with the captured tab? A change of source breaks that connection. The capturing application can detect that using Capture Handle, but what can it do then? Inform the user why some features suddenly disappear?!

The user is not interested in excuses.
The user is not interested in the distinction between application and browser.
The user protests: **"If I was not supposed to press that button - why did you ever show it to me?"**

### Solution
The browser should not guess whether a share-this-tab-instead button should be presented to the user. (Or any other UA-specific mechanism for switching sources.)

So how about:
```webidl
enum SourceChangePreferenceEnum {
  "include",
  "exclude"
};

dictionary DisplayMediaStreamConstraints {
  SourceChangePreferenceEnum sourceChange;
};
```

### Is this safe?
The application does NOT gain control over changing the source. It only opts into the user agent exposing to the user the option to change the source. The user remains in full control over whether the source is changed, what it is changed to, and when it is changed.

eladalon1983:
Fixes #223.

***
<a href="https://pr-preview.s3.amazonaws.com/eladalon1983/mediacapture-screen-share/pull/225.html" title="Last updated on Jun 16, 2022, 4:35 PM UTC (a7d1723)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/225/1eeb650...eladalon1983:a7d1723.html" title="Last updated on Jun 16, 2022, 4:35 PM UTC (a7d1723)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -738,7 +738,7 @@
               <tr>
                 <td>
                   <dfn id="idl-def-SelfCapturePreferenceEnum.include">include</dfn>
-            </td>
+                </td>
                 <td>
                   The application prefers the surface be included
                   among the choices offered.
--- a/index.html
+++ b/index.html
@@ -728,13 +728,13 @@
             };
           </pre>
           <table  data-dfn-for="SelfCapturePreferenceEnum" class="simple">
-	    <caption>{{SelfCapturePreferenceEnum}} Enumeration description</caption>
+      <caption>{{SelfCapturePreferenceEnum}} Enumeration description</caption>
             <thead>
               <tr>
-		<th>Enum value</th><th>Description</th>
+    <th>Enum value</th><th>Description</th>
               </tr>
-	    </thead>
-	    <tbody>
+      </thead>
+      <tbody>
               <tr>
                 <td>
                   <dfn id="idl-def-SelfCapturePreferenceEnum.include">include</dfn>
@@ -797,19 +797,19 @@
           </table>
         </section>
         <section>
-          <h2><dfn>DynamicSourceOptionPreferenceEnum</dfn></h2>
+          <h2><dfn>SurfaceSwitchingPreferenceEnum</dfn></h2>
           <p>
             Describes whether an application invoking {{MediaDevices/getDisplayMedia()}}
             would like the user agent to offer the user the option to dynamically change
             the source [=display surface=] during the capture.
           </p>
           <pre class="idl">
-            enum DynamicSourceOptionPreferenceEnum {
-              "include",
-              "exclude"
+            enum SurfaceSwitchingPreferenceEnum {
+              "yes",
+              "no"
             };
           </pre>
-          <table  data-dfn-for="DynamicSourceOptionPreferenceEnum" class="simple">
+          <table  data-dfn-for="SurfaceSwitchingPreferenceEnum" class="simple">
             <tbody>
               <tr>
                 <th colspan="2">
@@ -818,7 +818,7 @@
               </tr>
               <tr>
                 <td>
-                  <dfn id="idl-def-DynamicSourceOptionPreferenceEnum.include">include</dfn>
+                  <dfn id="idl-def-SurfaceSwitchingPreferenceEnum.yes">yes</dfn>
                 </td>
                 <td>
                   The application prefers that option to dynamically change the source
@@ -827,7 +827,7 @@
               </tr>
               <tr>
                 <td>
-                  <dfn id="idl-def-DynamicSourceOptionPreferenceEnum.exclude">exclude</dfn>
+                  <dfn id="idl-def-SurfaceSwitchingPreferenceEnum.no">no</dfn>
                 </td>
                 <td>
                   The application prefers that option to dynamically change the source
@@ -850,7 +850,7 @@ dictionary DisplayMediaStreamConstraints {
   (boolean or MediaTrackConstraints) audio = false;
   SelfCapturePreferenceEnum selfBrowserSurface;
   SystemAudioPreferenceEnum systemAudio;
-  DynamicSourceOptionPreferenceEnum dynamicSourceOption;
+  SurfaceSwitchingPreferenceEnum surfaceSwitching;
 };</pre>
             <section>
               <h2>Dictionary <a class="idlType">DisplayMediaStreamConstraints</a>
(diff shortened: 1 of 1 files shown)


eladalon1983:
Hello @jan-ivar and @youennf. I believe this is in-line with what we've discussed. Could you please review?

eladalon1983:
Friendly ping.

eladalon1983:
Friendly ping @jan-ivar and @youennf.

index.html
```diff
@@ -809,6 +850,7 @@ <h2>DisplayMediaStreamConstraints</h2>
   (boolean or MediaTrackConstraints) audio = false;
   SelfCapturePreferenceEnum selfBrowserSurface;
   SystemAudioPreferenceEnum systemAudio;
+  DynamicSourceOptionPreferenceEnum dynamicSourceOption;
```
alvestrand: I find the name cryptic. Can we name it canChangeCapturedSurface?

index.html
```diff
@@ -796,6 +796,47 @@ <h2><dfn>SystemAudioPreferenceEnum</dfn></h2>
...
+          <h2><dfn>DynamicSourceOptionPreferenceEnum</dfn></h2>
+          <p>
+            Describes whether an application invoking {{MediaDevices/getDisplayMedia()}}
+            would like the user agent to offer the user the option to dynamically change
+            the source [=display surface=] during the capture.
+          </p>
+          <pre class="idl">
+            enum DynamicSourceOptionPreferenceEnum {
+              "include",
+              "exclude"
```
alvestrand, line 809: I would prefer "allow" and "deny".

index.html
eladalon1983: I'm OK with whichever name @jan-ivar and @youennf will agree to. Including this one.

index.html
eladalon1983: I'd be happy to change this if Youenn and Jan-Ivar agree.

index.html
aboba: I like Harald's suggestion.

index.html
aboba: Agree that "Allow" and "Deny" is better than "include"/"exclude".

index.html
jan-ivar: A "change" may mean a change to the single souorce. How about source "switching"?

Regarding "allow" and "deny" those seem a poor match for what we've agreed is only a hint. 

How about `sourceSwitching: "prefer" | "disprefer"` ?

jan-ivar: `surfaceSwitching: "yes" | "no"`

jan-ivar: Actually, I like "include" and "exclude" here since it's consistent with recent enums we've added around surface inclusion/exclusion. This is slightly different since it's a dynamic feature to switch surfaces later, but seems similar in that it has to do with what surfaces the user can choose.

Those other enums were also hints, so being consistent in naming with them would be nice, and might help readers intuitively understand they're all hints

index.html
jan-ivar: Actually, I like "include" and "exclude" here since it's consistent with recent enums we've added around surface inclusion/exclusion. This is slightly different since it's a dynamic feature to switch surfaces later, but seems similar in that it has to do with what surfaces the user can choose.

Those other enums were also hints, so being consistent in naming with them would be nice, and might help readers intuitively understand they're all hints

index.html
```diff
@@ -796,6 +796,47 @@ <h2><dfn>SystemAudioPreferenceEnum</dfn></h2>
...
+                <th colspan="2">
+                  Enumeration description
+                </th>
+              </tr>
+              <tr>
+                <td>
+                  <dfn id="idl-def-SurfaceSwitchingPreferenceEnum.include">include</dfn>
+                </td>
+                <td>
+                  The application prefers that option to dynamically change the source
```
```suggestion
                  The application prefers that an option to dynamically switch the source
```

index.html
```diff
@@ -796,6 +796,47 @@ <h2><dfn>SystemAudioPreferenceEnum</dfn></h2>
...
+                  [=display surface=] during the capture be offered to the user.
+                </td>
+              </tr>
+              <tr>
+                <td>
+                  <dfn id="idl-def-SurfaceSwitchingPreferenceEnum.exclude">exclude</dfn>
+                </td>
+                <td>
+                  The application prefers that option to dynamically change the source
```
```suggestion
                  The application prefers that an option to dynamically switch the source
```

index.html
```diff
@@ -855,6 +897,14 @@ <h2>Dictionary <a class="idlType">DisplayMediaStreamConstraints</a>
                   If present, signals whether the application would like system audio
                   to be included among the possible audio sources offered to the user.
                 </dd>
+                <dt>
+                  <dfn><code>dynamicSourceOption</code></dfn> of type {{SurfaceSwitchingPreferenceEnum}}
+                </dt>
+                <dd>
+                  If present, signals whether the application would like the
+                  user agent to offer the user the option to dynamically change
```
```suggestion
                  user agent to offer the user an option to dynamically switch
```

jan-ivar (approved):

youennf (approved):

aboba (approved):

PR #235, opened 2022-09-29
Add CaptureController to the spec

eladalon1983:
***
<a href="https://pr-preview.s3.amazonaws.com/eladalon1983/mediacapture-screen-share/pull/235.html" title="Last updated on Sep 29, 2022, 6:55 PM UTC (14821a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/235/cd59841...eladalon1983:14821a5.html" title="Last updated on Sep 29, 2022, 6:55 PM UTC (14821a5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -215,12 +215,12 @@
               <li>
                 <p>
                   Let <var>controller</var> be <var>options</var>.<code>controller</code> if present,
-                  or null otherwise.
+                  or <code>null</code> otherwise.
                 </p>
               </li>
               <li>
                 <p>
-                  If controller is not null, run the following steps:
+                  If controller is not <code>null</code>, run the following steps:
                 </p>
                   <ol>
                     <li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -721,6 +743,38 @@ <h2>
...
+            to {{MediaDevices/getDisplayMedia()}} or its resulting stream or tracks.
+          </p>
+          <p>
+            Any given [=capture-session=] is associated with at most one {{CaptureController}}.
+          </p>
+          <p>
+            At most one {{CaptureController}} is associated with any given [=capture-session=].
+          </p>
+          <pre class="idl">
+            [Exposed=Window]
```
beaufortfrancois: As the "MediaDevices" interface is "SecureContext", I'd propose `CaptureController` to be `SecureContext` as well.

index.html
```diff
@@ -212,6 +212,27 @@ <h2>
                 <p>Let <var>options</var> be the method's first
                 argument.</p>
               </li>
+              <li>
+                <p>
+                  Let <var>controller</var> be <var>options</var>.<code>controller</code>.
```
beaufortfrancois:
```suggestion
                  Let <var>controller</var> be <var>options</var>.<code>controller</code> if present, or null otherwise.
```

index.html
```diff
@@ -212,6 +212,27 @@ <h2>
...
+                </p>
+              </li>
+              <li>
+                <p>
+                  If <var>controller</var> is <code>undefined</code>, continue to
```
beaufortfrancois: I'd go with this pattern. WDYT?

```
If controller is not null, run the following steps:
 - If controller.isBound is true, return a promise rejected with a
   DOMException object whose name attribute has the value InvalidStateError.
 - Set controller.isBound to true.
```

alvestrand (commented):
Agree with Beaufort's comments.
If this handle accumulates functionality, it is worth doing. At the moment it seems like a lot of stuff to go through for just one function, but let's try it.

index.html
eladalon1983: Done.

index.html
eladalon1983: Done.

index.html
eladalon1983: Done.

eladalon1983:
Merging based on `editors can integrate` from earlier today, now that all feedback has been incorporated. Improvements can be merged later.

Issue #230, opened 2022-08-31
A CaptureController object for getDisplayMedia() 


jan-ivar:
As promised in https://www.w3.org/2022/05/17-webrtc-minutes.html#t05 a dedicated controller object for getDisplayMedia.

An initial version would be aimed at solving #190 for now:
```js
const controller = new CaptureController();
const stream = await navigator.mediaDevices.getDisplayMedia({controller});

// Default is no focus when a controller is associated.
if (focus) {
  await controller.focus();
}
```

The controller is associated 1 ←→ 1 upon getDisplayMedia success, and cannot be re-associated after that.

1. Presence of the controller suppresses the automatic bring-to-front with focus that happens today.
2. Without a controller, getDisplayMedia would bring-to-front with focus like today for backwards compatibility.
3. Apps may call focus() anytime ahead of getDisplayMedia or right after it (upto 1 second after seems fine, but let’s iterate on details). Rejects with InvalidStateError after a timeout or if capture has stopped, whichever is sooner.
4. focus() resolves after the captured window or tab has been brought to front and focused, or asap for monitor display surfaces, but no earlier than gDM success.

youennf:
> The controller is associated 1 ←→ 1 upon getDisplayMedia success, and cannot be re-associated after that.

I do not see much value in reusing controller objects.
It would be simpler if the controller is associated upon getDisplayMedia being called (except in case of synchronous exception). That would cover the case of getDisplayMedia being called twice in a row for instance. First one would proceed, second one would reject synchronously.

> 1\. Presence of the controller suppresses the automatic bring-to-front with focus that happens today.

Seems fine given above caveat.

> 2\. Without a controller, getDisplayMedia would bring-to-front with focus like today for backwards compatibility.

This is UA territory, I am not sure the current spec says anything about this.
A note in the spec would probably be good here.

> 3\. Apps may call focus() anytime ahead of getDisplayMedia or right after it (upto 1 second after seems fine, but let’s iterate on details)

I am not a big fan of a timer here.
First we might enter UA territory, with different timeout values.
Second, if you leave something like 1 second, people might want to go to network (say with Capture Handle identity) to decide what to do. This might lead to races between the timer and networking.

Maybe we should start with something restrictive and relax to a timer approach if we see good use cases for it.

Or we could require an explicit decision from the capturer and delay video frame delivery until the decision is made.
focus() would then take a boolean parameter to express the decision of the capturer.
[...]

PR #240, opened 2022-10-05
Define CaptureController.setFocusBehavior()

### LINKED ISSUE #190: Conditional Focus (When Display-Capture Starts)
## Problem
When an application starts capturing a display-surface, the user agent faces a decision - should the captured display-surface be brought to the forefront, or should the capturing application retain focus.

The user agent is mostly agnostic of the nature of the capturing and captured applications, and is therefore ill-positioned to make an informed decision.

In contrast, the capturing application is familiar with its own properties, and is therefore better suited to make this decision. Moreover, by reading [displaySurface](https://www.w3.org/TR/screen-capture/#dom-mediatrackconstraintset-displaysurface) and/or using [Capture Handle](https://wicg.github.io/capture-handle/), the capturing application can learn about the captured display-surface, driving an even more informed decision.

## Sample Use Case
For **example**, a video conferencing application **may** wish to:
* Focus a captured application that users typically interact with during the call, like a text editor.
* Retain for itself focus when the captured display-surface is non-interactive content, like a video.
  * (Using [Capture Handle](https://wicg.github.io/capture-handle/), the capturing application may even allow the user to remotely start/pause the video.)

## Suggested Solution and Demo
* I have produced a [spec-draft](https://eladalon1983.github.io/conditional-focus/index.html) of my suggested solution.
* I have also implemented it behind a flag in Chrome and produced and a [demo](https://eladalon1983.github.io/conditional-focus/demo/).
  * Note that this demo works on Chrome Canary.
  * Further instructions are provided in the demo itself.

## Sample Code
```
const stream = await navigator.mediaDevices.getDisplayMedia();
track.focus(ShouldFocus(stream) ? "focus-captured-surface" : "no-focus-change");

function ShouldFocus(stream) {
  const [track] = stream.getVideoTracks();
  if (sampleUsesCaptureHandle) {
    // Assume logic discriminating focusability by origin,
    // for instance focusing anything except https://collaborator.com.
    const captureHandle = track.getCaptureHandle();
    return ShouldFocusOrigin(captureHandle && captureHandle.origin);
  } else {  // Assume Capture Handle is not a thing.
    // Assume the application is only interested in focusing tabs, not windows.
    return track.getSettings().displaySurface == 'browser';
  }
}
```

## Security Concerns
One noteworthy security concerns is that allowing switching focus at an arbitrary moment could allow clickjacking attacks. The [suggested spec](https://eladalon1983.github.io/conditional-focus/index.html) addresses this concern by limiting the time when focus-switching may be triggered/suppressed - the application may only decide about focus immediately[\*] upon the resolution of the `Promise<MediaStream>`. (See the [spec-draft](https://eladalon1983.github.io/conditional-focus/index.html) for more details about what "immediately" means and how I suggest various edge-cases be handled.)

eladalon1983:
Fixes #190.

***
<a href="https://pr-preview.s3.amazonaws.com/eladalon1983/mediacapture-screen-share/pull/240.html" title="Last updated on Oct 13, 2022, 3:18 PM UTC (d0d2ce8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/240/55359af...eladalon1983:d0d2ce8.html" title="Last updated on Oct 13, 2022, 3:18 PM UTC (d0d2ce8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -808,7 +808,7 @@
                 </th>
               </tr>
               <tr>
-                <td>
+                <td nowrap>
                   <dfn id="idl-def-CaptureStartFocusBehavior.focus-captured-surface">focus-captured-surface</dfn>
                 </td>
                 <td>
--- a/index.html
+++ b/index.html
@@ -974,7 +974,7 @@
                 </li>
                 <li>
                   <p>
-                    Run the [=finalize focus decision algorithm=].
+                    Run the [=finalize focus decision algorithm=] on [=this=].
                   </p>
                 </li>
               </ol>
--- a/index.html
+++ b/index.html
@@ -388,36 +388,39 @@
                       This invocation of {{MediaDevices/getDisplayMedia()}} is now
                       considered to have produced a new <dfn>capture-session</dfn>.
                     </p>
+                  </li>
+                  <li>
+                    If <var>controller</var> is not <code>null</code>:
+                    <ol>
+                      <li>
+                        <p>
+                          Set <var>controller</var>.{{CaptureController/[[videoTrack]]}}
+                          to <var>stream</var>'s video track.    
+                        </p>
+                      </li>
+                      <li>
+                        <p>
+                          <!-- TODO: Be more exact about which entity loses focus. -->
+                          At this point, it is guaranteed that
+                          <var>controller</var>.{{CaptureController/[[initialFocusLost]]}}
+                          is <code>false</code>. If TODO ever
+                          <a data-cite="!HTML/#focused">loses focus</a>, set it back
+                          to <code>true</code>.
+                        </p>
+                      </li>
+                      <li>
+                        <p>
+                          Schedule a task to set
+                          <var>controller</var>.{{CaptureController/[[focusWindowClosed]]}}
+                          to <code>true</code>.
+                        </p>
+                      </li>
+                    </ol>
+                  </li>
+                  <li>
                     <p>
-                      If <var>controller</var> is not <code>null</code>:
-                      <ul>
-                        <li>
-                          <p>
-                            Set <var>controller</var>.{{CaptureController/[[videoTrack]]}}
-                            to <var>stream</var>'s video track.    
-                          </p>
-                        </li>
-                        <li>
-                          <p>
-                            <!-- TODO: Be more exact about which entity loses focus. -->
-                            At this point, it is guaranteed that
-                            <var>controller</var>.{{CaptureController/[[initialFocusLost]]}}
-                            is <code>false</code>. If TODO ever
-                            <a data-cite="!HTML/#focused">loses focus</a>, set it back
-                            to <code>true</code>.
-                          </p>
-                        </li>
-                        <li>
-                          <p>
-                            <!-- TODO: Add a task to close the window. -->
-                            TODO
-                          </p>
-                        </li>
-                      </ul>
-                    </p>
-                    <p>
-                      <a>Resolve</a> <var>p</var> with <var>stream</var> and abort
-                      these steps. 
+                      <a>Resolve</a> <var>p</var> with <var>stream</var> and
+                      abort these steps. 
                     </p>
                   </li>
                   <li>
--- a/index.html
+++ b/index.html
@@ -901,11 +901,7 @@
                     <td><dfn>[[\FocusWindowClosed]]</dfn></td>
                     <td><code>false</code></td>
                     <td>
-                      Tracks explicit closure of the window of opportunity, which happens when
-                      {{CaptureController/setFocusBehavior()}} is called, or [=task to close the window of opportunity to set
-                      focus behavior | the task to close the window=] executes.
-                      As applications can easily avoid such errors, {{CaptureController/setFocusBehavior()}} throws 
-                      when this slot is set to <code>true</code>.
+                      Set to true once the opportunity to set focus behavior has passed.
                     </td>
                   </tr>
                 </tbody>
--- a/index.html
+++ b/index.html
@@ -808,7 +808,7 @@
                 </th>
               </tr>
               <tr>
-                <td nowrap>
+                <td>
                   <dfn id="idl-def-CaptureStartFocusBehavior.focus-captured-surface">focus-captured-surface</dfn>
                 </td>
                 <td>
--- a/index.html
+++ b/index.html
@@ -975,67 +975,67 @@
                 </li>
               </ol>
             </dd>
-            <p>
-              The <dfn>finalize focus decision algorithm</dfn>, given a
-              <var>controller</var>, consists of running
-              the following steps:
-            </p>
+          </dl>
+          <p>
+            The <dfn>finalize focus decision algorithm</dfn>, given a
+            <var>controller</var>, consists of running
+            the following steps:
+          </p>
+          <ol>
+            <li>
+              <p>
+                If <var>controller</var>.{{CaptureController/[[FocusDecisionFinalized]]}}
+                is <code>true</code>, abort these steps.
+              </p>
+            </li>
+            <li>
+              <p>
+                Set <var>controller</var>.{{CaptureController/[[FocusDecisionFinalized]]}}
+                to <code>true</code>.
+              </p>
+            </li>
+            <li>
+              <p>
+                If <var>controller</var>.{{CaptureController/[[FocusChangeDisabled]]}}
+                is <code>true</code>, abort these steps.
+              </p>
+            </li>
+            <li>
+              <p>
+                If too much time has elapsed since the [=capture-session=] started,
+                the user agent SHOULD set {{CaptureController/[[FocusDecisionFinalized]]}}
+                to <code>true</code>. The timespan is left up to the user agent,
+                but it is recommended that a value of one second be used.
+              </p>
+            </li>
+            <li>
+              <p>
+                If <var>controller</var>.{{CaptureController/[[DisplaySurfaceType]]}} is neither
+                {{DisplayCaptureSurfaceType/"browser"}} nor {{DisplayCaptureSurfaceType/"window"}},
+                abort these steps.
+              </p>
+            </li>
+            <li>
+              <p>
+                Run the following step  [=in parallel=]:
+              </p>
+            </li>
             <ol>
               <li>
                 <p>
-                  If <var>controller</var>.{{CaptureController/[[FocusDecisionFinalized]]}}
-                  is <code>true</code>, abort these steps.
+                  If <var>controller</var>.{{CaptureController/[[FocusBehavior]]}}
+                  is {{CaptureStartFocusBehavior/"no-focus-change"}},
+                  focus the capturing document.
                 </p>
-              </li>
-              <li>
                 <p>
-                  Set <var>controller</var>.{{CaptureController/[[FocusDecisionFinalized]]}}
-                  to <code>true</code>.
+                  If <var>controller</var>.{{CaptureController/[[FocusBehavior]]}}
+                  is {{CaptureStartFocusBehavior/"focus-captured-surface"}},
+                  focus the [=display surface=] referred to by
+                  <var>controller</var>.{{CaptureController/[[Source]]}}.
                 </p>
               </li>
-              <li>
-                <p>
-                  If <var>controller</var>.{{CaptureController/[[FocusChangeDisabled]]}}
-                  is <code>true</code>, abort these steps.
-                </p>
-              </li>
-              <li>
-                <p>
-                  If too much time has elapsed since the [=capture-session=] started,
-                  the user agent SHOULD set {{CaptureController/[[FocusDecisionFinalized]]}}
-                  to <code>true</code>. The timespan is left up to the user agent,
-                  but it is recommended that a value of one second be used.
-                </p>
-              </li>
-              <li>
-                <p>
-                  If <var>controller</var>.{{CaptureController/[[DisplaySurfaceType]]}} is neither
-                  {{DisplayCaptureSurfaceType/"browser"}} nor {{DisplayCaptureSurfaceType/"window"}},
-                  abort these steps.
-                </p>
-              </li>
-              <li>
-                <p>
-                  Run the following step  [=in parallel=]:
-                </p>
-              </li>
-              <ol>
-                <li>
-                  <p>
-                    If <var>controller</var>.{{CaptureController/[[FocusBehavior]]}}
-                    is {{CaptureStartFocusBehavior/"no-focus-change"}},
-                    focus the capturing document.
-                  </p>
-                  <p>
-                    If <var>controller</var>.{{CaptureController/[[FocusBehavior]]}}
-                    is {{CaptureStartFocusBehavior/"focus-captured-surface"}},
-                    focus the [=display surface=] referred to by
-                    <var>controller</var>.{{CaptureController/[[Source]]}}.
-                  </p>
-                </li>
-              </ol>
-            </ol>  
-          </dl>
+            </ol>
+          </ol>  
         </section>
         <section>
           <h2><dfn>SelfCapturePreferenceEnum</dfn></h2>
--- a/index.html
+++ b/index.html
@@ -853,61 +853,63 @@
               <dfn>constructor</dfn>
             </dt>
             <dd>
-              Run the following steps:
-              <ol>
-                <li>
-                  <p>
-                    Create one internal slot: <dfn data-dfn-for="CaptureController">[[\IsBound]]</dfn>,
-                    initialized to <code>false</code>.
-                    <div class="note">
-                      This internal slot tracks whether [=this=] has been associated with a [=capture-session=].
-                    </div>
-                  </p>
-                </li>
-                <li>
-                  <p>
-                    Create one internal slot: <dfn data-dfn-for="CaptureController">[[\Source]]</dfn>,
-                    initialized to <code>null</code>.
-                    <div class="note">
-                      This internal slot tracks with which source the [=capture-session=] is associated.
-                      It is distinct from {{CaptureController/[[IsBound]]}}, in that {{CaptureController/[[IsBound]]}}
-                      is set even if {{MediaDevices/getDisplayMedia()}} rejects, whereas {{CaptureController/[[Source]]}}
+              Create a new object with the following internal slots:
+              <table class="simple">
+                <thead>
+                  <tr>
+                    <th>Internal Slot</th>
+                    <th>Initial value</th>
+                    <th>Description (<em>non-normative</em>)</th>
+                  </tr>
+                </thead>
+                <tbody>
+                  <tr>
+                    <td><dfn>[[\IsBound]]</dfn></td>
+                    <td><code>false</code></td>
+                    <td>
+                      Tracks whether an application has attempted to associate [=this=]
+                      with a [=capture-session=], by providing it as a value to
+                      {{CaptureController/setFocusBehavior()}}.
+                    </td>
+                  </tr>
+                  <tr>
+                    <td><dfn>[[\Source]]</dfn></td>
+                    <td><code>null</code></td>
+                    <td>
+                      Tracks with which source the [=capture-session=] is associated.
+                      It is distinct from {{CaptureController/[[IsBound]]}}, in that
+                      {{CaptureController/[[IsBound]]}} is set even if
+                      {{MediaDevices/getDisplayMedia()}} rejects, whereas {{CaptureController/[[Source]]}}
                       will only be set if {{MediaDevices/getDisplayMedia()}} succeeds.
-                    </div>
-                  </p>
-                </li>
-                <li>
-                  <p>
-                    Create one internal slot: <dfn data-dfn-for="CaptureController">[[\DisplaySurfaceType]]</dfn>,
-                    initialized to <code>null</code>.
-                    <div class="note">
-                      Once capture starts, this will be set to the type of the capture [=display surface=].
-                    </div>
-                  </p>
-                </li>
-                <li>
-                  <p>
-                    Create one internal slot: <dfn data-dfn-for="CaptureController">[[\InitialFocusLost]]</dfn>,
-                    initialized to <code>false</code>.
-                    <div class="note">
-                      This internal slot ensures that calls to {{CaptureController/setFocusBehavior()}} fail
+                    </td>
+                  </tr>
+                  <tr>
+                    <td><dfn>[[\DisplaySurfaceType]]</dfn></td>
+                    <td><code>null</code></td>
+                    <td>
+                      Once capture starts, this will be set to the type of the captured [=display surface=].
+                    </td>
+                  </tr>
+                  <tr>
+                    <td><dfn>[[\InitialFocusLost]]</dfn></td>
+                    <td><code>false</code></td>
+                    <td>
+                      Ensures that calls to {{CaptureController/setFocusBehavior()}} fail
                       without throwing if the user manually focuses another tab or native application.
-                    </div>
-                  </p>
-                </li>
-                <li>
-                  <p>
-                    Create one internal slot: <dfn data-dfn-for="CaptureController">[[\FocusWindowClosed]]</dfn>,
-                    initialized to <code>false</code>.
-                    <div class="note">
-                      This internal slot tracks explicit closure of the window of opportunity, which happens when
+                    </td>
+                  </tr>
+                  <tr>
+                    <td><dfn>[[\FocusWindowClosed]]</dfn></td>
+                    <td><code>false</code></td>
+                    <td>
+                      Tracks explicit closure of the window of opportunity, which happens when
                       {{CaptureController/setFocusBehavior()}} is called, or the task to close the window executes.
                       As applications can easily avoid such errors, {{CaptureController/setFocusBehavior()}} throws 
                       when this slot is set to <code>true</code>.
-                    </div>
-                  </p>
-                </li>
-              </ol>
+                    </td>
+                  </tr>
+                </tbody>
+              </table>
             </dd>
             <dt>
               <dfn>setFocusBehavior</dfn>
--- a/index.html
+++ b/index.html
@@ -220,7 +220,7 @@
               </li>
               <li>
                 <p>
-                  If controller is not <code>null</code>, run the following steps:
+                  If <var>controller</var> is not <code>null</code>, run the following steps:
                 </p>
                   <ol>
                     <li>
@@ -384,9 +384,31 @@
                     <var>message</var>)</code>.</p>
                   </li>
                   <li>
-                    <p><a>Resolve</a> <var>p</var> with <var>stream</var> and
-                    abort these steps. This invocation of {{MediaDevices/getDisplayMedia()}}
-                    is now considered to have produced a new <dfn>capture-session</dfn></p>
+                    <p>
+                      This invocation of {{MediaDevices/getDisplayMedia()}} is now
+                      considered to have produced a new <dfn>capture-session</dfn>.
+                    </p>
+                    <p>
+                      If <var>controller</var> is not <code>null</code>:
+                      <ul>
+                        <li>
+                          Set <var>controller</var>.{{CaptureController/[[videoTrack]]}}
+                          to <var>stream</var>'s video track.    
+                        </li>
+                        <li>
+                          <!-- TODO: Be more exact about which entity loses focus. -->
+                          At this point, it is guaranteed that
+                          <var>controller</var>.{{CaptureController/[[initialFocusLost]]}}
+                          is <code>false</code>. If TODO ever
+                          <a data-cite="!HTML/#focused">loses focus</a>, set it back
+                          to <code>true</code>.
+                        </li>
+                      </ul>
+                    </p>
+                    <p>
+                      <a>Resolve</a> <var>p</var> with <var>stream</var> and abort
+                      these steps. 
+                    </p>
                   </li>
                   <li>
                     <p><em>Permission Failure</em>: [=Reject=]
@@ -745,6 +767,46 @@
           </div>
         </section>
         <section>
+          <h2><dfn>CaptureStartFocusBehavior</dfn></h2>
+          <p>
+            Describes whether an application invoking {{CaptureController/setFocusBehavior()}}
+            would like the user agent to focus the [=display surface=] associated with
+            that {{CaptureController}}'s [=capture-session=].
+          </p>
+          <pre class="idl">
+            enum CaptureStartFocusBehavior {
+              "focus-captured-surface",
+              "no-focus-change"
+            };    
+          </pre>
+          <table  data-dfn-for="CaptureStartFocusBehavior" class="simple">
+            <tbody>
+              <tr>
+                <th colspan="2">
+                  Enumeration description
+                </th>
+              </tr>
+              <tr>
+                <td>
+                  <dfn id="idl-def-CaptureStartFocusBehavior.focus-captured-surface">focus-captured-surface</dfn>
+                </td>
+                <td>
+                  The application prefers that the [=display surface=] associated with
+                  this {{CaptureController}}'s [=capture-session=] be focused.
+                </td>
+              </tr>
+              <tr>
+                <td>
+                  <dfn id="idl-def-CaptureStartFocusBehavior.no-focus-change">no-focus-change</dfn>
+                </td>
+                <td>
+                  The application prefers that the user agent not change focus.
+                </td>
+              </tr>
+            </tbody>
+          </table>
+        </section>
+        <section>
           <h2><dfn>CaptureController</dfn></h2>
           <p>
             A {{CaptureController}} object may be associated with a [=capture-session=].
@@ -762,7 +824,7 @@
             [Exposed=Window, SecureContext]
             interface CaptureController {
               constructor();
-              // TODO: Add setFocusBehavior() in a separate PR.
+              undefined setFocusBehavior(CaptureStartFocusBehavior focusBehavior);
             };
           </pre>
           <dl data-link-for="CaptureController" data-dfn-for="CaptureController" class="methods">
@@ -770,10 +832,118 @@
               <dfn>constructor</dfn>
             </dt>
             <dd>
+              Run the following steps:
+              <ol>
+                <li>
+                  <p>
+                    Create one internal slot: <dfn data-dfn-for="CaptureController">[[\isBound]]</dfn>,
+                    initialized to <code>false</code>.
+                    <div class="note">
+                      This internal slot tracks whether [=this=] has been associated with a [=capture-session=].
+                    </div>
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    Create one internal slot: <dfn data-dfn-for="CaptureController">[[\videoTrack]]</dfn>,
+                    initialized to <code>null</code>.
+                    <div class="note">
+                      <!-- TODO: Ensure surfaceSwitching closes the focus window. -->
+                      This internal slot tracks with which [=display surface=] the [=capture-session=] is associated.
+                      It is distinct from {{CaptureController/[[isBound]]}} in that {{CaptureController/[[isBound]]}}
+                      is set even if {{MediaDevices/getDisplayMedia()}} rejects.
+                    </div>
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    Create one internal slot: <dfn data-dfn-for="CaptureController">[[\initialFocusLost]]</dfn>,
+                    initialized to <code>false</code>.
+                    <div class="note">
+                      This internal slot ensures that calls to {{CaptureController/setFocusBehavior()}} fail
+                      without throwing if the user manually focuses another tab or native application.
+                    </div>
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    Create one internal slot: <dfn data-dfn-for="CaptureController">[[\focusWindowClosed]]</dfn>,
+                    initialized to <code>false</code>.
+                    <div class="note">
+                      This internal slot tracks explicit closure of the window of opportunity, which happens when
+                      {{CaptureController/setFocusBehavior()}} is called, or the task to close the window executes.
+                      As applications can easily avoid such errors, {{CaptureController/setFocusBehavior()}} when
+                      this slot is set to <code>true</code> throw.
+                    </div>
+                  </p>
+                </li>
+              </ol>
+            </dd>
+            <dt>
+              <dfn>setFocusBehavior</dfn>
+            </dt>
+            <dd>
               <p>
-                Create one internal slot: <dfn data-dfn-for="CaptureController">[[\isBound]]</dfn>,
-                initialized to <code>false</code>.
+                Run the following algorithm:
               </p>
+              <ol>
+                <li>
+                  <p>
+                    If {{CaptureController/[[isBound]]}} is <code>false</code>,
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[videoTrack]]}} is <code>null</code>,
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[videoTrack]]}}.{{MediaStreamTrack/getSettings()}}.{{displaySurface}}
+                    is neither {{DisplayCaptureSurfaceType/"browser"}} nor {{DisplayCaptureSurfaceType/"window"}},
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    <!-- TODO: Fix the ReadyState reference. -->
+                    If {{CaptureController/[[videoTrack]]}}'s
+                    <code>ReadyState</code> is NOT {{MediaStreamTrackState/"live"}},
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[focusWindowClosed]]}} is <code>true</code>,
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[initialFocusLost]]}} is <code>true</code>,
+                    abort these steps.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    The user agent SHOULD abort these steps if too much time has elapsed since
+                    the [=capture-session=] started. The timespan is left up to the user agent,
+                    but it is recommended that a value of one second be used.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    Set {{CaptureController/[[focusWindowClosed]]}} to <code>true</code>.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    Focus the [=display surface=] referred to by {{CaptureController/[[videoTrack]]}}.
+                  </p>
+                </li>
+              </ol>
             </dd>
           </section>
         <section>
--- a/index.html
+++ b/index.html
@@ -939,7 +939,7 @@
                 </li>
                 <li>
                   <p>
-                    If the {{CaptureController/[[Source]]}} has been
+                    If {{CaptureController/[[Source]]}} has been
                     <a data-cite="GETUSERMEDIA#source-stopped">stopped</a>,
                     [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
                   </p>
--- a/index.html
+++ b/index.html
@@ -1034,7 +1034,8 @@
                 </li>
               </ol>
             </ol>  
-          </section>
+          </dl>
+        </section>
         <section>
           <h2><dfn>SelfCapturePreferenceEnum</dfn></h2>
           <p>
@@ -1755,7 +1756,7 @@ dictionary DisplayMediaStreamOptions {
         </p>
         <p>
           When capturing a [= display surface/window =] or other <a>display surface</a> that is partially transparent,
-          any content behind it will not be captured</a>.
+          any content behind it will not be captured.
         </p>
         <p>
           There is a risk that the user prompt be exposed to the web page for a short amount of time
--- a/index.html
+++ b/index.html
@@ -392,16 +392,26 @@
                       If <var>controller</var> is not <code>null</code>:
                       <ul>
                         <li>
-                          Set <var>controller</var>.{{CaptureController/[[videoTrack]]}}
-                          to <var>stream</var>'s video track.    
+                          <p>
+                            Set <var>controller</var>.{{CaptureController/[[videoTrack]]}}
+                            to <var>stream</var>'s video track.    
+                          </p>
                         </li>
                         <li>
-                          <!-- TODO: Be more exact about which entity loses focus. -->
-                          At this point, it is guaranteed that
-                          <var>controller</var>.{{CaptureController/[[initialFocusLost]]}}
-                          is <code>false</code>. If TODO ever
-                          <a data-cite="!HTML/#focused">loses focus</a>, set it back
-                          to <code>true</code>.
+                          <p>
+                            <!-- TODO: Be more exact about which entity loses focus. -->
+                            At this point, it is guaranteed that
+                            <var>controller</var>.{{CaptureController/[[initialFocusLost]]}}
+                            is <code>false</code>. If TODO ever
+                            <a data-cite="!HTML/#focused">loses focus</a>, set it back
+                            to <code>true</code>.
+                          </p>
+                        </li>
+                        <li>
+                          <p>
+                            <!-- TODO: Add a task to close the window. -->
+                            TODO
+                          </p>
                         </li>
                       </ul>
                     </p>
--- a/index.html
+++ b/index.html
@@ -903,6 +903,11 @@
               <ol>
                 <li>
                   <p>
+                    Let <var>focusBehavior</var> be the method's first argument.
+                  </p>
+                </li>
+                <li>
+                  <p>
                     If {{CaptureController/[[isBound]]}} is <code>false</code>,
                     [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
                   </p>
@@ -950,6 +955,12 @@
                 <li>
                   <p>
                     Set {{CaptureController/[[focusWindowClosed]]}} to <code>true</code>.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If <var>focusBehavior</var> is NOT {{CaptureStartFocusBehavior/"focus-captured-surface"}},
+                    abort these steps.
                   </p>
                 </li>
                 <li>
--- a/index.html
+++ b/index.html
@@ -903,7 +903,7 @@
               <ol>
                 <li>
                   <p>
-                    If {{CaptureController/[[isBound]]}} is <code>true</code>,
+                    If {{CaptureController/[[isBound]]}} is <code>false</code>,
                     [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
                   </p>
                 </li>
--- a/index.html
+++ b/index.html
@@ -887,8 +887,8 @@
                     <div class="note">
                       This internal slot tracks explicit closure of the window of opportunity, which happens when
                       {{CaptureController/setFocusBehavior()}} is called, or the task to close the window executes.
-                      As applications can easily avoid such errors, {{CaptureController/setFocusBehavior()}} when
-                      this slot is set to <code>true</code> throw.
+                      As applications can easily avoid such errors, {{CaptureController/setFocusBehavior()}} throws 
+                      when this slot is set to <code>true</code>.
                     </div>
                   </p>
                 </li>
--- a/index.html
+++ b/index.html
@@ -236,14 +236,6 @@
                       Set <var>controller</var>.{{CaptureController/[[IsBound]]}} to <code>true</code>.
                     </p>
                   </li>
-                  <li>
-                    <p>
-                      If the <a>current settings object</a>'s [=relevant global object=]'s
-                      [=associated `Document`=] ever <a data-cite="!HTML/#focused">loses focus</a>,
-                      set <var>controller</var>.{{CaptureController/[[InitialFocusLost]]}}
-                      to <code>true</code>.
-                    </p>
-                  </li>
                 </ol>
               </li>
               <li>
@@ -423,6 +415,14 @@
                           to <code>true</code> when it executes.
                         </p>
                       </li>
+                      <li>
+                        <p>
+                          If the <a>current settings object</a>'s [=relevant global object=]'s
+                          [=associated `Document`=] ever <a data-cite="!HTML/#focused">loses focus</a>,
+                          set <var>controller</var>.{{CaptureController/[[InitialFocusLost]]}}
+                          to <code>true</code>.
+                        </p>
+                      </li>    
                     </ol>
                   </li>
                   <li>
--- a/index.html
+++ b/index.html
@@ -417,7 +417,8 @@
                       </li>
                       <li>
                         <p>
-                          Queue a task to run the [=finalize focus decision algorithm=].
+                          Queue a task to run the [=finalize focus decision algorithm=]
+                          on <var>controller<var>.
                         </p>
                       </li>
                     </ol>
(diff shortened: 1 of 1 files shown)


eladalon1983:
Still some work here, but early feedback welcome.

index.html
```diff
@@ -384,9 +384,45 @@ <h2>
...
+                      considered to have produced a new <dfn>capture-session</dfn>.
+                    </p>
+                  </li>
+                  <li>
+                    If <var>controller</var> is not <code>null</code>:
+                    <ol>
+                      <li>
+                        <p>
+                          Set <var>controller</var>.{{CaptureController/[[videoTrack]]}}
+                          to <var>stream</var>'s video track.
```
youennf: It makes more sense for me to tie the controller to the source itself instead of the track.

index.html
```diff
@@ -384,9 +384,45 @@ <h2>
...
+                    <ol>
+                      <li>
+                        <p>
+                          Set <var>controller</var>.{{CaptureController/[[videoTrack]]}}
+                          to <var>stream</var>'s video track.    
+                        </p>
+                      </li>
+                      <li>
+                        <p>
+                          <!-- TODO: Be more exact about which entity loses focus. -->
```
youennf: I wonder how this works if focus is already lost as part of the window selection step.

index.html
```diff
@@ -762,18 +838,126 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                    is neither {{DisplayCaptureSurfaceType/"browser"}} nor {{DisplayCaptureSurfaceType/"window"}},
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    <!-- TODO: Fix the ReadyState reference. -->
+                    If {{CaptureController/[[videoTrack]]}}'s
+                    <code>ReadyState</code> is NOT {{MediaStreamTrackState/"live"}},
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
```
youennf: An application could clone the track, stop the original track but the source would still be live with the cloned track.
It seems best to rely on the source and verify here whether the source is live or not.

index.html
```diff
@@ -762,18 +838,126 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[focusWindowClosed]]}} is <code>true</code>,
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[initialFocusLost]]}} is <code>true</code>,
```
youennf: At this step, we might want to run steps in parallel, given this might actually happen in another process.

index.html
```diff
@@ -762,18 +838,126 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    Set {{CaptureController/[[focusWindowClosed]]}} to <code>true</code>.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    Focus the [=display surface=] referred to by {{CaptureController/[[videoTrack]]}}.
```
youennf: It seems focus should take the parameter of setFocusBehavior as input.

index.html
```diff
@@ -762,18 +838,126 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                  <p>
+                    If {{CaptureController/[[videoTrack]]}}.{{MediaStreamTrack/getSettings()}}.{{displaySurface}}
+                    is neither {{DisplayCaptureSurfaceType/"browser"}} nor {{DisplayCaptureSurfaceType/"window"}},
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    <!-- TODO: Fix the ReadyState reference. -->
+                    If {{CaptureController/[[videoTrack]]}}'s
```
youennf: A CaptureController might not have a videoTrack/source at the time setFocusBehavior is called.
In that case, it should probably store the preference somewhere so that it can be used at the time the source is created and where UA might try to set focus.

youennf (commented):
Some comments below.
In general, it seems the spec assumes that the UA is focused at the time of resolution, which might not be the case if the captured window is already focused due to the window selection process.
Also, where is the code stating that if controller is not null, automatic focusing rule is prevented?

index.html
eladalon1983: Generally yes, but there's the technical issue of detecting when the source changes through [surfaceSwitching](https://github.com/w3c/mediacapture-screen-share/pull/225/files) and closing the window when that happens. I intend to tackle this with a follow-up PR. I think this will likely work better if we specify a handle on the track rather than the source.

eladalon1983: (Also, capture-actions might follow to the next source, so they might still want to be tied to the track, to facilitate "track's current source.")

index.html
eladalon1983: > 6. If [[[initialFocusLost]]]() is true, abort these steps.

I believe this should work for Safari?
Do you want to call it out more clearly, or phrase it differently?

eladalon1983:
> In general, it seems the spec assumes that the UA is focused at the time of resolution, which might not be the case if the captured window is already focused due to the window selection process.

There is a line that should handle this: "If [[initialFocusLost]] is true, abort these steps."
I've referred to it in an in-line comment, I assume we'll continue there.

> Also, where is the code stating that if controller is not null, automatic focusing rule is prevented?

The TPAC meeting concluded with [this](https://github.com/w3c/mediacapture-screen-share/issues/230#issuecomment-1245839419) result: "We'll not specify the default behavior (focus/no-focus) for either case (controller/no-controller)."

(That comment was published before the minutes came out, but I first checked with Dom that the summary matched the minutes he was going to publish.)

index.html
eladalon1983: I believe you mean this sequence of calls:
```js
const captureController = new CaptureController;
captureController.setFocusBehavior("no-focus-change");
navigator.mediaDevices({video: true, controller: captureController});
```

My reading was that there was no consensus for it. (Namely, I am fine with it, but I thought Jan-Ivar was not interested.) So I'm not supporting it at the moment.

I propose that we tackle this as a follow-up even if we do have consensus. It seems easy to add later.

index.html
eladalon1983: Done.

index.html
eladalon1983: (Follow-up PR, I mean; not necessarily follow-up discussion. From my side, I do not foresee myself either insisting or objecting. I'm fine either way.)

index.html
eladalon1983: I see now that I had a faulty assumption baked in of `initialFocusLost` being false at a relatively late point. I've moved the rule to track focus changes to an earlier part of the algorithm, before the window selection can commence. Does that work?

index.html
eladalon1983: I think the principle of least surprise would still instruct us to treat focus the way we'd treat a stopped track with no clones. It seems to me unlikely that something else would be expected/desired by an application that (1) supplies a controller and (2) immediately stops the original track as soon as gDM's promise resolves.

One exception to this, perhaps, would be track transfer. Perhaps that's best handled by calling out and handling transfer explicitly. (And if so - I propose this should be done in a follow-up PR.)

Wdyt?

index.html
eladalon1983: To be sure I understand you correctly, could you clarify the proposed change?

index.html
```diff
@@ -762,18 +839,137 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                    <code>ReadyState</code> is NOT {{MediaStreamTrackState/"live"}},
```
beaufortfrancois: How about checking first if videoTrack's ReadyState is NOT "live" before videoTrack.getSettings().displaySurface? In other words, switching "4." and "5.".

index.html
```diff
@@ -762,18 +839,137 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    Set {{CaptureController/[[focusWindowClosed]]}} to <code>true</code>.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If <var>focusBehavior</var> is NOT {{CaptureStartFocusBehavior/"focus-captured-surface"}},
```
beaufortfrancois: Shouldn't we specify that the display surface MUST NOT be focused  when "no-focus-change". Or at least a MAY NOT be focused.

index.html
```diff
@@ -762,18 +839,137 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                </li>
+                <li>
+                  <p>
+                    Create one internal slot: <dfn data-dfn-for="CaptureController">[[\focusWindowClosed]]</dfn>,
+                    initialized to <code>false</code>.
+                    <div class="note">
+                      This internal slot tracks explicit closure of the window of opportunity, which happens when
+                      {{CaptureController/setFocusBehavior()}} is called, or the task to close the window executes.
+                      As applications can easily avoid such errors, {{CaptureController/setFocusBehavior()}} when
+                      this slot is set to <code>true</code> throw.
```
beaufortfrancois: I found it hard to read.

```suggestion
                      As applications can easily avoid such errors, {{CaptureController/setFocusBehavior()}} throws 
                      when this slot is set to <code>true</code>.
```

index.html
eladalon1983: What's the rationale?

index.html
eladalon1983: I don't want to specify what other APIs may or may not do. This API surface switches focus if the application requests it, and is no-op otherwise.

index.html
beaufortfrancois: https://www.w3.org/TR/mediacapture-streams/#dom-mediastreamtrack-getsettings algorithm already checks if track's ReadyState is "ended". I thought we could spare one check.

eladalon1983: I don't understand how that saves a step. `getSettings()` checks the readyState so that it may produce the same values from before ending. It does not bail early or return anything different when that happens. Or have I misread it?

Also, I'd not want to build a dependency on the specifics of `getSettings()`. That seems brittle.

beaufortfrancois: Note that I don't feel strongly about this.

> Also, I'd not want to build a dependency on the specifics of getSettings(). That seems brittle.

Agree.

eladalon1983: I don't freely strongly that the order should be as it currently is, either. I'm happy to change it if you can explain why the inverse order is better.

beaufortfrancois: We're good. Feel free to close/resolve this conversation.

eladalon1983: Cool.

index.html
eladalon1983: Following the editors' meeting yesterday, this PR has now switched to relying on the source.

index.html
```diff
@@ -744,6 +787,46 @@ <h2>
...
+          </pre>
+          <table  data-dfn-for="CaptureStartFocusBehavior" class="simple">
+            <tbody>
+              <tr>
+                <th colspan="2">
+                  Enumeration description
+                </th>
+              </tr>
+              <tr>
+                <td>
```
beaufortfrancois:
```suggestion
                <td nowrap>
```

After
image
Before
image

index.html
```diff
@@ -762,18 +845,145 @@ <h2><dfn>CaptureController</dfn></h2>
...
               constructor();
-              // TODO: Add setFocusBehavior() in a separate PR.
+              undefined setFocusBehavior(CaptureStartFocusBehavior focusBehavior);
             };
           </pre>
           <dl data-link-for="CaptureController" data-dfn-for="CaptureController" class="methods">
             <dt>
               <dfn>constructor</dfn>
             </dt>
+            <dd>
```
beaufortfrancois, line 864: Instead of constructor steps, you may want to use a table that describes all internal slots as in https://w3c.github.io/web-nfc/#the-ndefreader-object

image

index.html
```diff
@@ -762,18 +845,145 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                    </div>
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    Create one internal slot: <dfn data-dfn-for="CaptureController">[[\FocusWindowClosed]]</dfn>,
+                    initialized to <code>false</code>.
+                    <div class="note">
+                      This internal slot tracks explicit closure of the window of opportunity, which happens when
+                      {{CaptureController/setFocusBehavior()}} is called, or the task to close the window executes.
```
beaufortfrancois: Can we link "the task to close the window" to the "Schedule a task to..."?

index.html
```diff
@@ -220,18 +220,30 @@ <h2>
...
+                  </li>
+                  <li>
+                    <p>
+                      Set <var>controller</var>.{{CaptureController/[[IsBound]]}} to <code>true</code>.
+                    </p>
+                  </li>
+                  <li>
+                    <p>
+                      If the <a>current settings object</a>'s [=relevant global object=]'s
+                      [=associated `Document`=] ever <a data-cite="!HTML/#focused">loses focus</a>,
```
beaufortfrancois: Shall we consider [page visibility](https://html.spec.whatwg.org/multipage/interaction.html#page-visibility) instead of document focus?

eladalon1983: getDisplayMedia uses this condition.
> If the [current settings object](https://html.spec.whatwg.org/multipage/webappapis.html#current-settings-object)'s [relevant global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global)'s [associated Document](https://html.spec.whatwg.org/multipage/window-object.html#concept-document-window) is NOT [fully active](https://html.spec.whatwg.org/multipage/browsers.html#fully-active) or does NOT [have focus](https://html.spec.whatwg.org/multipage/#gains-focus), return a promise [rejected](https://webidl.spec.whatwg.org/#reject) with a [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException) object whose [name](https://webidl.spec.whatwg.org/#dom-domexception-name) attribute has the value [InvalidStateError](https://webidl.spec.whatwg.org/#invalidstateerror).

index.html
eladalon1983: Looks good. Done.

index.html
eladalon1983: Sure, that does make it easier for a reader to find their way around the doc.

eladalon1983:
@jan-ivar and @youennf, this PR is ready for review.

index.html
eladalon1983: Actually, that seems to fail the bots.

beaufortfrancois: It says to use CSS. I'd suggest you replace `nowrap` with `style="white-space:nowrap"` to keep this behavior.

eladalon1983: That worked.

index.html
```diff
@@ -762,18 +845,148 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                    <td>
+                      Once capture starts, this will be set to the type of the captured [=display surface=].
+                    </td>
+                  </tr>
+                  <tr>
+                    <td><dfn>[[\InitialFocusLost]]</dfn></td>
+                    <td><code>false</code></td>
+                    <td>
+                      Ensures that calls to {{CaptureController/setFocusBehavior()}} fail
+                      without throwing if the user manually focuses another tab or native application.
```
beaufortfrancois: This raises two questions:

1. When `getDisplayMedia()` is called, the user can switch tabs but the browser prompt will still be shown until user make a decision. By monitoring focus changes **before** `getDisplayMedia` returned promise is resolved, we would allow a situation where `setFocusBehavior` would silently fail if user focused another tab while the browser prompt was displayed. The solution would be to monitor focus changes **after** `getDisplayMedia` returned promise is resolved.
2. Even with that suggested change, I think `InitialFocusLost` is not required in the spec as the "focus change" JS event would not be fired before `setFocusBehavior` be called. Am I understand this right?

What are your thoughts?

eladalon1983: * The question of whether the picker should persist when the user tabs away-then-back is orthogonal.
* The PR did not make it sufficiently clear that `InitialFocusLost` should only start tracking focus after the user makes their choice. That's now fixed.
* Events go on a queue. I prefer to skip it and ensure `setFocusBehavior` is rendered no-op immediately, so that a user would never find their `alt-tab` invalidated.

dontcallmedom:
the CI error should be ignored for now; I'm not sure yet if it's the result of https://github.com/w3c/respec-web-services/issues/353 or the fact that our charter was only very recently renewed.

eladalon1983:
> the CI error should be ignored for now; I'm not sure yet if it's the result of [w3c/respec-web-services#353](https://github.com/w3c/respec-web-services/issues/353) or the fact that our charter was only very recently renewed.

Thanks

index.html
```diff
@@ -225,18 +225,22 @@ <h2>
...
                       the value {{InvalidStateError}}.
-                    </li>
-                    <li>
-                      Set <var>controller</var>.{{CaptureController/[[isBound]]}} to <code>true</code>.
-                    </li>
+                    </p>
+                  </li>
+                  <li>
+                    <p>
+                      Set <var>controller</var>.{{CaptureController/[[IsBound]]}} to <code>true</code>.
```
jan-ivar, line 225: (Not new with this PR but), right now we seem to bind the controller in the middle of our input and state validation steps.

E.g. if we don't have transient activation we don't bind, but on `{video: false}` we do? Seems arbitrary.

It might be better to do the binding last or first. Since there's always going to be WebIDL tests first, how do we feel about making the binding the last step before the in-parallel steps?

We should be able to write a WPT test for that using `Promise.race` on the rejected promise.

index.html
```diff
@@ -389,9 +393,48 @@ <h2>
...
+                        <p>
+                          Set <var>controller</var>.{{CaptureController/[[DisplaySurfaceType]]}} to the 
+                          to <var>stream</var>'s video track's {{DisplayCaptureSurfaceType}}.
+                        </p>
+                      </li>
+                      <li>
+                        <p>
+                          Schedule a <dfn>task to close the window of opportunity to set focus behavior</dfn>,
+                          which will set <var>controller</var>.{{CaptureController/[[FocusWindowClosed]]}}
+                          to <code>true</code> when it executes.
```
jan-ivar: We need to use "[queue a task](https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task)" language here. This spec has languished a bit, and we should update it at some point to match modern specs more (e.g. what task source are we using?), but small steps.

I wouldn't make this step into a definition since it doesn't seem to support a normative reference from another algorithm, nor does it seem like a well-defined concept.

I like "window of opportunity", but worry the metaphors may be a bit on the nose and confusing in this context. How about `[[FocusOpportunityExpired]]`?

```suggestion
                          Queue a task to set <var>controller</var>.{{CaptureController/[[FocusOpportunityExpired]]}}
                          to <code>true</code>.
```

index.html
```diff
@@ -389,9 +393,48 @@ <h2>
...
+                        </p>
+                      </li>
+                      <li>
+                        <p>
+                          If the <a>current settings object</a>'s [=relevant global object=]'s
+                          [=associated `Document`=] ever <a data-cite="!HTML/#focused">loses focus</a>,
```
jan-ivar: This step doesn't belong here among the other sequential steps, and needs to be broken out as its own paragraph outside of this algorithm I think, since it happens later, not here.

Also, it's probably the top-level document we want, even though this language is wrong elsewhere as well today until we fix https://github.com/w3c/mediacapture-main/issues/752. If an iframe looses focus we don't want to call this.

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                    <th>Initial value</th>
+                    <th>Description (<em>non-normative</em>)</th>
+                  </tr>
+                </thead>
+                <tbody>
+                  <tr>
+                    <td><dfn>[[\IsBound]]</dfn></td>
+                    <td><code>false</code></td>
+                    <td>
+                      Tracks whether an application has attempted to associate [=this=]
```
```suggestion
                      Whether an application has attempted to associate [=this=]
```

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                      with a [=capture-session=], by providing it as a value to
+                      {{CaptureController/setFocusBehavior()}}.
```
```suggestion
                      with a [=capture-session=].
```

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                    </td>
+                  </tr>
+                  <tr>
+                    <td><dfn>[[\Source]]</dfn></td>
+                    <td><code>null</code></td>
+                    <td>
+                      Tracks with which source the [=capture-session=] is associated.
```
```suggestion
                      The source of the associated [=capture-session=].
```

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                      It is distinct from {{CaptureController/[[IsBound]]}}, in that
+                      {{CaptureController/[[IsBound]]}} is set even if
+                      {{MediaDevices/getDisplayMedia()}} rejects, whereas {{CaptureController/[[Source]]}}
+                      will only be set if {{MediaDevices/getDisplayMedia()}} succeeds.
```
jan-ivar: I'd focus on describing what the state represents, and leave algorithms to explain how the state is used, to avoid saying things twice. They appear distinct enough to me.
```suggestion
```

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                    <td>
+                      Once capture starts, this will be set to the type of the captured [=display surface=].
+                    </td>
+                  </tr>
+                  <tr>
+                    <td><dfn>[[\InitialFocusLost]]</dfn></td>
+                    <td><code>false</code></td>
+                    <td>
+                      Ensures that calls to {{CaptureController/setFocusBehavior()}} fail
+                      without throwing if the user manually focuses another tab or native application.
```
jan-ivar: Again I'd focus inward on describing what the state represents, without describing what algorithms accomplish with it (outward). E.g. something like "Becomes true if the top-level document loses focus". People can cross-link to figure out how state ends up being used in algorithms (they may even have cross-linked to get here and find such descriptions circular).

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
-              // TODO: Add setFocusBehavior() in a separate PR.
+              undefined setFocusBehavior(CaptureStartFocusBehavior focusBehavior);
             };
           </pre>
           <dl data-link-for="CaptureController" data-dfn-for="CaptureController" class="methods">
             <dt>
               <dfn>constructor</dfn>
             </dt>
+            <dd>
+              Create a new object with the following internal slots:
```
jan-ivar: I like this simpler format, but we need to mention the type
```suggestion
              Create a new {{CaptureController}} object with the following internal slots:
```

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                  <tr>
+                    <td><dfn>[[\InitialFocusLost]]</dfn></td>
+                    <td><code>false</code></td>
+                    <td>
+                      Ensures that calls to {{CaptureController/setFocusBehavior()}} fail
+                      without throwing if the user manually focuses another tab or native application.
+                    </td>
+                  </tr>
+                  <tr>
+                    <td><dfn>[[\FocusWindowClosed]]</dfn></td>
```
jan-ivar: FocusWindow sounds like it could be a thing. How about:
```suggestion
                    <td><dfn>[[\FocusOpportunityExpired]]</dfn></td>
```

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                    <td><code>false</code></td>
+                    <td>
+                      Tracks explicit closure of the window of opportunity, which happens when
+                      {{CaptureController/setFocusBehavior()}} is called, or [=task to close the window of opportunity to set
+                      focus behavior | the task to close the window=] executes.
+                      As applications can easily avoid such errors, {{CaptureController/setFocusBehavior()}} throws 
+                      when this slot is set to <code>true</code>.
```
jan-ivar: Again I'd prefer something inward and simpler, like
```suggestion
                      Set to true once the opportunity to set focus behavior has passed.
```

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+              <ol>
+                <li>
+                  <p>
+                    Let <var>focusBehavior</var> be the method's first argument.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[IsBound]]}} is <code>false</code>,
+                    [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
```
jan-ivar: Why disallow calling it ahead of time?

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[Source]]}} is <code>null</code>,
+                    [=exception/throw=] a "{{InvalidStateError}}" {{DOMException}}.
```
jan-ivar: When can [[Source]] be null?

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[Source]]}} has been
+                    <a data-cite="GETUSERMEDIA#source-stopped">stopped</a>,
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
```
jan-ivar: Don't we reserve `NotAllowedError` for permission errors? This sounds more like a `InvalidStateError`.

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                  <p>
+                    If {{CaptureController/[[DisplaySurfaceType]]}} is neither
+                    {{DisplayCaptureSurfaceType/"browser"}} nor {{DisplayCaptureSurfaceType/"window"}},
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[FocusWindowClosed]]}} is <code>true</code>,
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
```
jan-ivar: Maybe `InvalidStateError`?

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                    If {{CaptureController/[[Source]]}} has been
+                    <a data-cite="GETUSERMEDIA#source-stopped">stopped</a>,
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[DisplaySurfaceType]]}} is neither
+                    {{DisplayCaptureSurfaceType/"browser"}} nor {{DisplayCaptureSurfaceType/"window"}},
+                    [=exception/throw=] a "{{NotAllowedError}}" {{DOMException}}.
```
jan-ivar: Same here. InvalidStateError sounds more correct to me.

index.html
```diff
@@ -767,18 +850,155 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    Run the following steps in parallel:
+                  </p>
+                  <ol>
+                    <li>
+                      <p>
+                        Focus the [=display surface=] referred to by {{CaptureController/[[Source]]}}.
```
jan-ivar: Do we want to mandate something more specific here?

If {{CaptureController/[[DisplaySurfaceType]]}} is `"window"`, do we focus the window? In case that window belongs to the user agent, does it focus the visible document or not? Or do we not want to say?

If {{CaptureController/[[DisplaySurfaceType]]}} is `"browser"`, do we focus the top-level browsing context associated with {{CaptureController/[[Source]]}}'s [=display surface=]?

jan-ivar (commented):
Thanks for tackling this! [[Source]] looks right!

index.html
eladalon1983: I've made the requested change.

index.html
eladalon1983: I'm fine with `[[FocusOpportunityExpired]]`.

index.html
eladalon1983: I think are specs can be very inaccessible to newcomers, and verbosity in non-normative sections can help with that. But we can circle back to that later; right now, I am focused on getting the normative section merged. Suggestion committed.

index.html
eladalon1983: If you pass `CaptureController` to a `getDisplayMedia()` call that does not culminate with the user actually selecting something to capture.

index.html
eladalon1983: Because our discussions seemed to suggest reduced scope had consensus, with the following journey disallowed:
```js
const captureController = new CaptureController;
captureController.setFocusBehavior("no-focus-change");
navigator.mediaDevices({video: true, controller: captureController});
```
If we're all interested in allowing, then I am happy to introduce this as a follow-up.

index.html
eladalon1983: Done.

index.html
eladalon1983: Done.

index.html
eladalon1983: Done.

index.html
eladalon1983: Let's start by leaving it unspecified and letting our respective browsers do what they do today. We can discuss aligning the finer details as a potential follow-up.

index.html
eladalon1983: If you could suggest a good place to move it, I'm happy to oblige. But note that we need to somehow clarify that this condition only comes into effect *after* the user makes their selection. That is, we don't count as `InitialFocusLost` cases were the user interacts with another tab, native app, or OS-level widget, if that happens **before** they finish their interaction with the picker. This is why this step was put here originally. Do you have an idea for how to express that better?

index.html
eladalon1983: Moved it further up; now it's first.

eladalon1983:
Thanks for reviewing, @jan-ivar. I've responded to all threads. The call is back in your court.

index.html
jan-ivar: I am interested in calling the method ahead of time. I'm also wondering why the method isn't idempotent?

I'm curious, why did we reject a setter instead of a method? E.g.

```js
const captureController = new CaptureController;
captureController.focusBehavior = "no-focus-change";
navigator.mediaDevices({video: true, controller: captureController});
```
To clarify, I'm a fan of using the right JS surface where applicable. I was a fan of method when it was a simple `focus()` method, and I consider "set-methods" a bit of an anti-pattern really. But if both of you have good reasons for that that you can share, I'm willing to compromise.

index.html
jan-ivar: Yeah that language is a bit broken, which is https://github.com/w3c/mediacapture-main/issues/752.

index.html
eladalon1983: **For posterity:** These are responses to a comment which @jan-ivar moved to another thread.

> I am interested in calling the method ahead of time.

Okay, then let's add that as a follow-up PR, which I'll produce and upload for review after the current PR gets merged.

> I'm also wondering why the method isn't idempotent?

Because I don't expect this to be called more than once by a well-structured app, given how it can only be called within a very narrow window of opportunity. But the follow-up will change things so that we can call arbitrarily many times before gDM is called, with same/different values, and up to one time after gDM is called.

> I'm curious, why did we reject a setter instead of a method?

It's more natural for a method to (1) have side-effect and (2) throw when used at an inappropriate time.

index.html
eladalon1983: Let me know if you have a fix in mind that's uncontroversial enough to make it into this PR. Otherwise, I expect we'll be fixing as a follow-up. (I'm resolving this conversation for now, but that's a reversible action.)

index.html
jan-ivar: I see, yeah I wouldn't infer that solely from its placement here. I would place it right after these algorithm steps as a new paragraph with something like:

"When the top-level document loses focus, run the following steps on all CaptureController objects in that document and in documents of its nested browsing contexts:

1. If [[Source]] is null, abort these steps.
2. Set [[InitialFocusLost]] to true.

Would that work?

index.html
jan-ivar: I am interested in calling the method ahead of time. I'm also wondering why the method isn't idempotent?

I'm curious, why did we reject a setter instead of a method? E.g.

```js
const captureController = new CaptureController;
captureController.focusBehavior = "no-focus-change";
navigator.mediaDevices({video: true, controller: captureController});
```
To clarify, I'm a fan of using the right JS surface where applicable. I was a fan of method when it was a simple `focus()` method, and I consider "set-methods" a bit of an anti-pattern really. But if both of you have good reasons for that that you can share, I'm willing to compromise.

jan-ivar: (Note I moved this comment down to make it more visible)

index.html
jan-ivar: I see, if we want to allow calling it early then we would need to change this.

jan-ivar:
I think if we want to allow calling this method ahead of time we should probably nail that down right away, so we don't end up with something else shipped that might cause a web compat problem.

index.html
jan-ivar: But how does this work with UAs that defocus ahead of the prompt (like Chrome does?)

eladalon1983: Suggestion adopted.

jan-ivar: I wonder if we're in user agent territory here since some browsers are focused in this case and others aren't. What's the behavior we're protecting here? Is the idea that we don't want to compete with a user-driven focus change that might have happened in the interim?

index.html
eladalon1983: > (Note I moved this comment down to make it more visible)

Where's my response? Could you move it down here too? Feel free to edit this message.
Edit: I've done so myself.

eladalon1983:
> I think if we want to allow calling this method ahead of time we should probably nail that down right away, so we don't end up with something else shipped that might cause a web compat problem.

I don't think we would end up with a compat issue either way, but if you and Youenn can be responsive, I can expand the scope of this PR to also tackle this use-case too.

index.html
```diff
@@ -225,20 +215,34 @@ <h2>
...
+              <li>
+                <p>If the [=relevant global object=] of [=this=] does not have
+          [=transient activation=], return a promise <a>rejected</a>
+                with a {{DOMException}} object whose {{DOMException/name}}
+                attribute has the value {{InvalidStateError}}.</p>
+              </li>
+              <li>
+                <p>Let <var>options</var> be the method's first
+                argument.</p>
+              </li>
```
jan-ivar: Sorry this isn't the part that I suggested moving. Please move it back. I meant instead we could move the controller binding steps to right before the in parallel steps. Sorry for any confusion.

eladalon1983: I don't recall moving this. That line 238 is mis-indented indicates something went wrong along the way. Let's see about the fix...

index.html
eladalon1983: Indeed.

index.html
jan-ivar: A setter still allows us to throw exceptions, the only difference is we would have to add a `"default"` value. My interest here is that this seems like a better pattern that helps inform design questions (at least for me) like:

1. should this be settable ahead of time? yes
2. should it be idempotent? yes
3. is there a third inferred option implied by not calling it? yes (today's UA focus behavior)

@youennf WDYT?

index.html
eladalon1983: Having looked at it again - I'm not following. This looks correctly placed. Please look again. (And note that it's unrelated to the request to move the focus-tracking part to the bottom.)

index.html
eladalon1983: * With a setter, we can both set ahead of time as well as keep it idempotent.
* With a setter, we don't have to specify the default value.
* With a setter, we don't allow reading the value, which could be confusing to some apps, since this is a hint they set, and does not report back what the UA will actually do.
* With a setter, side-effects and throwing are more naturally expected.

eladalon1983: Btw, these are points which were in my original message. I don't think it's good to move your arguments around, it makes it hard to see my counter-arguments, which are left back in the original thread.

eladalon1983: I've hunted down my original responses too. Here:

> I am interested in calling the method ahead of time.

Okay, then let's add that as a follow-up PR, which I'll produce and upload for review after the current PR gets merged.
Edit: Outdated assuming that we can get quick turn-around on increasing scope.

> I'm also wondering why the method isn't idempotent?

Because I don't expect this to be called more than once by a well-structured app, given how it can only be called within a very narrow window of opportunity. But the follow-up will change things so that we can call arbitrarily many times before gDM is called, with same/different values, and up to one time after gDM is called.

> I'm curious, why did we reject a setter instead of a method?

It's more natural for a method to (1) have side-effect and (2) throw when used at an inappropriate time.

eladalon1983:
I've added the ability to call `setFocusBehavior()` ahead of the call to `getDisplayMedia()`. PTAL.

index.html
```diff
@@ -207,16 +207,6 @@ <h2>
             method is called, the user agent MUST run the following
             steps:</p>
             <ol>
-              <li>
```
eladalon1983, line 204: Note that the diff-tool just thinks this was removed. In effect, I just put something ahead of it and then fixed the indentation.

eladalon1983: @jan-ivar

index.html
```diff
@@ -767,19 +859,175 @@ <h2><dfn>CaptureController</dfn></h2>
...
+              <li>
+                <p>
+                  If {{CaptureController/[[FocusBehavior]]}} is NOT
+                  {{CaptureStartFocusBehavior/"focus-captured-surface"}},
+                  abort these steps.
+                </p>
+              </li>
+              <li>
+                <p>
+                  Run the following steps in parallel:
```
alvestrand: Is "Run the following steps" just using the same language as elsewhere, or is there a plan for more steps?
If it's just standard language, it should be "the following step" since there's just one.

index.html
```diff
@@ -767,19 +859,175 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                </p>
+              </li>
+              <ol>
+                <li>
+                  <p>
+                    Focus the [=display surface=] referred to by {{CaptureController/[[Source]]}}.
```
alvestrand: Is there a reference for the "Focus" action? (probably an HTML5 one)

alvestrand (approved):
Looks good to me.

index.html
jan-ivar: I see, the controller association moved up. However, [as I mentioned](https://github.com/w3c/mediacapture-screen-share/pull/240#discussion_r991648452) this still isn't going to guarantee the controller is always associated on failure e.g. if the WebIDL binding code throws TypeErrors on the constructor arguments. Are we OK with that?

Would it be more logical to move it down instead, after input validation, just ahead of the in-parallel steps, since it has side-effects? That way, if the returned promise is pending (as determined with `Promise.race`), the app can know the binding happened or not (something we can WPT test).

index.html
```diff
@@ -767,19 +859,175 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                    If {{CaptureController/[[Source]]}} has been
+                    <a data-cite="GETUSERMEDIA#source-stopped">stopped</a>,
+                    [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[DisplaySurfaceType]]}} is neither
+                    {{DisplayCaptureSurfaceType/"browser"}} nor {{DisplayCaptureSurfaceType/"window"}},
+                    [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
```
jan-ivar, line 957: First thought: this seems like an error of type, not state. Maybe `TypeError` and move it up one step?

Second thought: the "type" varies by what the user chooses in the picker... should we abort silently instead? Seems a footgun otherwise: Developer tests their code without error, but forgot to manually test choosing "Entire desktop" in the prompt which will throw in production.

Also, "Entire desktop" is already "in focus' so to speak, so why not call this success instead of failure?

index.html
```diff
@@ -389,9 +395,38 @@ <h2>
...
+                      </li>
+                      <li>
+                        <p>
+                          Set <var>controller</var>.{{CaptureController/[[DisplaySurfaceType]]}} to the 
+                          to <var>stream</var>'s video track's {{DisplayCaptureSurfaceType}}.
+                        </p>
+                      </li>
+                      <li>
+                        <p>
+                          Queue a task to run the [=finalize focus decision algorithm=].
```
```suggestion
                          Queue a task to run the [=finalize focus decision algorithm=]
                          on <var>controller<var>.
```

index.html
```diff
@@ -767,19 +859,175 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                  <p>
+                    If too much time has elapsed since the [=capture-session=] started,
+                    the user agent SHOULD set {{CaptureController/[[FocusDecisionFinalized]]}}
+                    to <code>true</code>. The timespan is left up to the user agent,
+                    but it is recommended that a value of one second be used.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    Run the [=finalize focus decision algorithm=].
```
```suggestion
                    Run the [=finalize focus decision algorithm=] on [=this=].
```

index.html
```diff
@@ -767,19 +859,175 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                    [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If too much time has elapsed since the [=capture-session=] started,
+                    the user agent SHOULD set {{CaptureController/[[FocusDecisionFinalized]]}}
+                    to <code>true</code>. The timespan is left up to the user agent,
+                    but it is recommended that a value of one second be used.
+                  </p>
```
jan-ivar: This part needs to move to the "finalize focus decision algorithm", or it won't catch all busy-waiting exploits.

index.html
```diff
@@ -767,19 +859,175 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                </li>
+                <li>
+                  <p>
+                    Run the [=finalize focus decision algorithm=].
+                  </p>
+                </li>
+              </ol>
             </dd>
+            <p>
+              The <dfn>finalize focus decision algorithm</dfn> consists of running
```
```suggestion
              The <dfn>finalize focus decision algorithm</dfn>, given a
              <var>controller</var>, consists of running
``````

Also, every reference to an internal slot {{CaptureController/[[Foo]]}} should be <var>controller</var>.{{CaptureController/[[Foo]]}}.

File: index.html [hunk 41]
```diff
@@ -767,19 +859,175 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                <li>
+                  <p>
+                    Let <var>focusBehavior</var> be the method's first argument.
+                  </p>
+                </li>
+                <li>
+                  <p>
+                    If {{CaptureController/[[Source]]}} is <code>null</code>,
+                    set {{CaptureController/[[FocusBehavior]]}} to <var>focusBehavior</var>
+                    and abort these steps.
```
jan-ivar, line 943: [[FocusBehavior]] must be set when [[Source]] is _not_ null as well, otherwise `focusBehavior` is unused.

index.html
```diff
@@ -767,19 +859,175 @@ <h2><dfn>CaptureController</dfn></h2>
...
+                </p>
+              </li>
+              <li>
+                <p>
+                  Set {{CaptureController/[[FocusDecisionFinalized]]}} to <code>true</code>.
+                </p>
+              </li>
+              <li>
+                <p>
+                  If {{CaptureController/[[InitialFocusLost]]}} is <code>true</code>,
```
jan-ivar: As mentioned by others, [[InitialFocusLost]] seems to assume UAs have focus while the picker is up. Does this hold?

It's a bit confusing. When I [test Chrome](https://jsfiddle.net/jib1/1vjpow8b/) the _page_ is blurred before the picker opens, but maybe the Chrome window itself retains focus at this point? This seems potentially UA dependent.

What situation exactly is this designed to protect? And do we need to figure it out in version 1?

Would [[DisqualifyingFocusChange]] be a better name until we figure out how it's supposed to work? Or maybe we leave it out until then?

index.html
```suggestion
                  Run the following steps [=in parallel=]:
```

index.html
jan-ivar: > It's more natural for a method to (1) have side-effect and (2) throw when used at an inappropriate time.

I don't think so. `video.srcObject = stream` has lots of side-effects, can throw, and is susceptible to timing (active/inactive streams).

> With a setter, we don't allow reading the value, which could be confusing to some apps, since this is a hint they set, and does not report back what the UA will actually do.

Note that by "setter" I mean attribute setter steps, the way the term is used in [e.g. HTML](https://html.spec.whatwg.org/multipage/interaction.html#dom-hidden), i.e. not a method. Thus there's necessarily a getter as well.

I don't think an attribute would be confusing. `controller.focusBehavior` could default to `"default"`, and be setable up to focus finalization steps, after which attempting to set it could produce `InvalidStateError`. At that point the UA could even fix the getter to return either `"focus-captured-surface"` (if focus was decided) or `"no-focus-change"` (if not) so the app can learn whether it was successful (which it currently cannot, even if setFocusBehavior() succeeds). 

That way, if we ever decide we want an event to fire when this has happened, we have the getter we'd need to abide by the [§ 7.7. Use plain Events for state](https://w3ctag.github.io/design-principles/#state-and-subclassing) design principle.

index.html
jan-ivar: The display surface may be a non-browser/non-HTML window, so I doubt it.

Once we flesh out details more (as I suggested in https://github.com/w3c/mediacapture-screen-share/pull/240#discussion_r991713555), the [system focus](https://html.spec.whatwg.org/multipage/interaction.html#tlbc-system-focus) in HTML might be something we could use when we're capturing `"browser"` surfaces (assuming we mean to give the document focus as well, and not just the browser window).

jan-ivar (commented):
Thanks for the changes! This looks promising. Still some more nits.

I'm also still curious whether we'd consider making it an attribute, as I think much of the algorithms could remain the same, and it'd open up a way to let apps know what happened.
 
Does that seem appealing to anyone? FWIW it's not the same discussion we had over the `focus()` promise, since this would only reveal the decision, not the timing of the actual focusing.

youennf:
The PR language is still tied to a particular selection picker model, which is not the only model available.
It seems useful to make this PR work for all know models, including when capture surface focus might be done as part of the surface selection.
The API would be a no-op as currently written in that case:
- Initial focus would be lost as per the current rule.
- If the web application states that it wants to retain focus, the UA should make an explicit action to refocus the web application, which is not specified by the finalise focus decision algorithm (step 5 would abort and step 6 is only about focusing the captured surface).

eladalon1983:
> The PR language is still tied to a particular selection picker model, which is not the only model available. It seems useful to make this PR work for all know models, including when capture surface focus might be done as part of the surface selection. The API would be a no-op as currently written in that case:
> 
> * Initial focus would be lost as per the current rule.
> * If the web application states that it wants to retain focus, the UA should make an explicit action to refocus the web application, which is not specified by the finalise focus decision algorithm (step 5 would abort and step 6 is only about focusing the captured surface).

@youennf, I have intentionally shaped the PR to be no-op for your case. Namely, at no point does it say that the user agent should avoid switching focus, leaving itself the flexibility to do so for reasons other than the app's request. If that's not sufficient, could you please suggest a concrete change that'd improve the shape in the way you seek?

index.html
eladalon1983: Done.

index.html
eladalon1983: Friendly ping @youennf for this question.

index.html
eladalon1983: Now that Jan-Ivar has migrated this discussion and I have copied over my responses, I am resolving this thread.

index.html
eladalon1983: > What situation exactly is this designed to protect?

The possibility that the user `alt-tabs` elsewhere, or a new native window pops up and steals focus.

> And do we need to figure it out in version 1?
> [...]
> Or maybe we leave it out until then?

Definitely not a must. Are you also in favor of dropping this from v1, @youennf?

index.html
eladalon1983: Done.

alvestrand:
Re setFocusBehavior() vs an attribute:

I think the setFocusBehavior() is better ergonomically - the mental model fits better with the API model - but technically, the only difference is that you get a getter for free if you make it an attribute.
To me, it makes little sense to want to look at how focus was set once setting focus has finished - it is only of interest while you are in the process of deciding on focus.

index.html
eladalon1983: > Note that by "setter" I mean attribute setter steps, the way the term is used in [e.g. HTML](https://html.spec.whatwg.org/multipage/interaction.html#dom-hidden), i.e. not a method. Thus there's necessarily a getter as well.

I know.

I also don't like that the getter would have to return something, while the UA default behavior might be different for tabs/windows, and so the UA might not actually be able to clarify its default behavior by returning a value. That's one more reason why I prefer a method (set-only) rather than a property (set+get).

index.html
eladalon1983: I think the mental model for the developer should be kept as simple as possible. "I can call setFocusBehavior" all the way up to when gDM is called, then exactly one more time, at latest when gDM's promise resolves" - this is already at the limit. I don't want to add the caveat of "...unless gDM fails, in which case I can call it again." As a developer, if your gDM call fails, try again from scratch.

eladalon1983: I think the mental model for the developer should be kept as simple as possible. `"I can call setFocusBehavior() multiple times before gDM is called, then exactly one more time, at latest when gDM's promise resolves"` - this is already at the limit. I don't want to add the caveat of `"...unless gDM fails, in which case I can call it again."` 

As a developer, if your gDM call fails, try again from scratch.

It's a shame that failures over WebIDL bindings don't get captured here, but I see it as an edge-case we can avoid worrying about. I don't think programs would typically run into this and then try to resolve such an error on the fly, programmatically. Rather, it will be a bug that someone will manually fix.

index.html
eladalon1983: Done.

index.html
eladalon1983: > Are we OK with that?

I've commented on that elsewhere. The comment threads are getting out of hand. I propose that we merge this and iterate. If there are substantive issues to resolve first, like methods vs. attribute, let's focus on them and leave uncontroversial spec-language nits for a follow-up.

eladalon1983: > Second thought: the "type" varies by what the user chooses in the picker... should we abort silently instead?

We do abort silently if `setFocusBehavior()` was called before gDM, when the surface type was unknown. If the call is made after the user makes their choice, failing to check the surface type is a Web-dev bug and should be fixed. By failing silently we would be delaying the time when the Web-dev would realize their mistake.

eladalon1983: > Also, "Entire desktop" is already "in focus' so to speak, so why not call this success instead of failure?

That depends on the OS and whether it's the current desktop or not.
At any rate, I really think this is all fine to discuss over v2.

youennf:
> @youennf, I have intentionally shaped the PR to be no-op for your case.

This makes this whole API useless for these UAs, meaning it will not be implemented.
It also casts doubts on the actual API design if it can only accommodate one single selection picker model.

> could you please suggest a concrete change

How about the following:
- Rename [InitialFocusLost]] to [[Disabled]] or [[FocusChangeDisabled]] for clarity.
- Make the controller listen to loss of page focus at the time the promise is resolved. If page already lost focus at this point, the controller remains active except if the page regains focus and loses it again.
- Mention that UAs may decide to set [[Disabled]] based on other heuristics.
-  Update the finalise focus decision algorithm to call focus on either the capturing surface or the captured surface based on [[FocusBehavior]].

index.html
eladalon1983: What scenario would not be caught? Once the UA is executing `setFocusBehavior()`,  the app can not busy-wait until `setFocusBehavior()` returns.

eladalon1983: Done.

index.html
eladalon1983: The line that sets it immediately after must have been dropped. I've fixed back to:
...
2) If [[[Source]]](https://github.com/w3c/mediacapture-screen-share/pull/240#dfn-source) is null, set [[[FocusBehavior]]](https://github.com/w3c/mediacapture-screen-share/pull/240#dfn-focusbehavior) to focusBehavior and abort these steps.
...
6) Set [[[FocusBehavior]]](https://github.com/w3c/mediacapture-screen-share/pull/240#dfn-focusbehavior) to focusBehavior.
...

eladalon1983:
The threads are really getting hard to follow, **and** the dog ate my homework, in that I have penned a response to one of Jan-Ivar's comments, that I now see was lost - unclear how. Namely, I want to point out that a property involves a getter that is problematic on many accounts, one **additional** one being that it's unclear what "default" means (recall that the default could be different for tabs/windows, and it's too early to know what the user chose). So before setting, the value is meaningless, and after setting, a getter is not necessary.

I propose that we:
1. Concentrate on aligning the substantive issues. Method vs. property, callability before gDM, etc.
2. Merge this PR.
3. Iterate on spec language.

I think the size and number of the threads make it too difficult to proceed otherwise.

eladalon1983:
> (@youennf) How about the following:
> ...

Thank you for the concrete proposal. I am happy to adopt a variant of this. However, I am not the only one who will have opinions, and this can take some time (the PR currently adds ~250 lines, and has roughly that many comments). This does not seem to change behavior, only improve the spec, so it should not be a blocker. I can commit to producing the follow-up PR by end-of-day of the day when we merge this PR.

youennf:
> so it should not be a blocker

I think we should demonstrate in this PR that the API is expected to work with more than Chrome current device picker.
This honestly does not seem like a lot of editorial work to achieve this.

eladalon1983:
> This honestly does not seem like a lot of editorial work to achieve this.

The challenge is in the excessive number of open comment threads in case your editorial vision does not immediately align with Jan-Ivar's.

eladalon1983:
So, I've tried to accommodate this one more round of spec-shape comments. Hopefully we can agree that this is enough for v1, or something very close to it.

> * Rename [InitialFocusLost]] to [[Disabled]] or [[FocusChangeDisabled]] for clarity.

I've gone with [[FocusChangeDisabled]] because [[Disabled]] would refer to future APIs too, and we don't intend that atm.

> * Make the controller listen to loss of page focus at the time the promise is resolved. If page already lost focus at this point, the controller remains active except if the page regains focus and loses it again.

That's already in the spec - see section with "When the top-level document loses focus, run the following steps on all {{CaptureController}} objects", noting that it's no-op before calling gDM(controller).

> * Mention that UAs may decide to set [[Disabled]] based on other heuristics.

I've added that.

> * Update the finalise focus decision algorithm to call focus on either the capturing surface or the captured surface based on [[FocusBehavior]].

I don't think focusing the capturing surface is desirable. It could interfere with the users own actions, and cannot otherwise help.

alvestrand:
@youennf claims that a 2-line change to the PR will address his concern. @eladalon1983 and @youennf will work to finish today.

index.html
```diff
@@ -399,6 +435,23 @@ <h2>
...
             </ol>
+            <p>
+              When the top-level document loses focus, run the following steps on all
+              {{CaptureController}} objects in that document and in documents of its nested
+              [=browsing contexts=]:
+            </p>
+            <ol>
+              <li>
+                <p>
+                  If {{CaptureController/[[IsBound]]}} is <code>false</code>, abort these steps.
```
youennf:
```suggestion
                  If {{CaptureController/[[Source]]}} is <code>undefined</code>, abort these steps.
```

youennf: Otherwise, we end up in the issue I mentionned previously:
- web app call gdm.
- user unfocus page (controller gets disabled).
- user focus back page and selects the surface in picker.

What we want to protect is focus change between the time the surface is picked and the time setFocus is called.

eladalon1983:
This is ready. Wdys @youennf?

index.html
```diff
@@ -761,20 +854,191 @@ <h2><dfn>CaptureController</dfn></h2>
...
+              <p>
+                Run the following step  [=in parallel=]:
+              </p>
+            </li>
+            <ol>
+              <li>
+                <p>
+                  If <var>controller</var>.{{CaptureController/[[FocusBehavior]]}}
+                  is {{CaptureStartFocusBehavior/"no-focus-change"}},
+                  focus the capturing document.
```
youennf:
```suggestion
                  focus the [=display surface=] representing the capturing document.
```

youennf (approved):

eladalon1983:
I see GitHub has had a hiccup when committing the last suggestions, and things somehow got merged without them. I'll fix that later tonight.

eladalon1983:
@beaufortfrancois, you've mentioned some commits were dropped, but as far as I can see these are in gh-pages. Please advise. (Note that https://github.com/w3c/mediacapture-screen-share/pull/245/files does not appear to be the PR to have fixed this; Youenn's suggestions appear on the left side, too.)

PR #254, opened 2022-11-04
Add missing suppressLocalAudioPlayback setting

beaufortfrancois:
I've noticed `suppressLocalAudioPlayback` setting was missing in https://w3c.github.io/mediacapture-screen-share/#extensions-to-mediatracksettings while https://w3c.github.io/mediacapture-screen-share/#dfn-suppresslocalaudioplayback says: 
> As a setting, this value indicates whether or not the user agent is applying local audio playback suppression to the source.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/mediacapture-screen-share/pull/254.html" title="Last updated on Nov 14, 2022, 8:36 AM UTC (231cb3d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/254/dc44099...beaufortfrancois:231cb3d.html" title="Last updated on Nov 14, 2022, 8:36 AM UTC (231cb3d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1431,8 +1431,8 @@ dictionary DisplayMediaStreamOptions {
             </dt>
             <dd>
               <p>
-                Indicates whether the <a href="#dfn-suppresslocalaudioplayback">suppressLocalAudioPlayback</a>
-                constraint is applied (<code>true</code>) or not (<code>false</code>).
+                Indicates whether or not the user agent is applying
+                [=local audio playback suppression=] to the source.
               </p>
             </dd>
           </dl>
--- a/index.html
+++ b/index.html
@@ -1425,6 +1425,15 @@ dictionary DisplayMediaStreamOptions {
                 constraint is applied (<code>true</code>) or not (<code>false</code>).
               </p>
             </dd>
+            <dt>
+               <dfn>suppressLocalAudioPlayback</dfn> of type {{boolean}}
+            </dt>
+            <dd>
+              <p>
+                Indicates whether the <a href="#dfn-suppressLocalAudioPlayback">suppressLocalAudioPlayback</a>
+                constraint is applied (<code>true</code>) or not (<code>false</code>).
+              </p>
+            </dd>
           </dl>
         </section>
         <section>
--- a/index.html
+++ b/index.html
@@ -701,8 +701,8 @@
               <td><dfn class=export>suppressLocalAudioPlayback</dfn></td>
               <td>{{ConstrainBoolean}}</td>
               <td>
-                <p>As a setting, this value indicates whether or not the constraint
-                for <a>local audio playback suppression</a> is applied to
+                <p>As a setting, this value indicates whether or not the user
+                agent is applying <a>local audio playback suppression</a> to
                 the source.</p>
                 <p>As a constraint, this value is only meaningful if the user
                 selects capturing a [= display surface/browser=] [=display surface=]. In that
--- a/index.html
+++ b/index.html
@@ -1431,7 +1431,7 @@ dictionary DisplayMediaStreamOptions {
             </dt>
             <dd>
               <p>
-                Indicates whether or not the user agent is applying
+                Indicates whether or not the application instructed the user agent to apply
                 [=local audio playback suppression=] to the source.
               </p>
             </dd>
--- a/index.html
+++ b/index.html
@@ -701,9 +701,9 @@
               <td><dfn class=export>suppressLocalAudioPlayback</dfn></td>
               <td>{{ConstrainBoolean}}</td>
               <td>
-                <p>As a setting, this value indicates whether or not the user
-                agent is applying <a>local audio playback suppression</a> to
-                the source.</p>
+                <p>As a setting, this value indicates whether or not the application
+                instructed the user agent to apply <a>local audio playback suppression</a>
+                to the source.</p>
                 <p>As a constraint, this value is only meaningful if the user
                 selects capturing a [= display surface/browser=] [=display surface=]. In that
                 case, a value of <code>true</code> indicates that the user
--- a/index.html
+++ b/index.html
@@ -701,8 +701,8 @@
               <td><dfn class=export>suppressLocalAudioPlayback</dfn></td>
               <td>{{ConstrainBoolean}}</td>
               <td>
-                <p>As a setting, this value indicates whether or not the user
-                agent is applying <a>local audio playback suppression</a> to
+                <p>As a setting, this value indicates whether or not the constraint
+                for <a>local audio playback suppression</a> is applied to
                 the source.</p>
                 <p>As a constraint, this value is only meaningful if the user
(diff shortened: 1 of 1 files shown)


eladalon1983:
Could you modify the text to indicate that the value reflects what the app applied, and not whether the browser complied? That is, consider two distinct tabs A and B capturing C. If A suppresses C's playback and B does not, then we should still return false for B's query; it must not be allowed to observe the existence of A and its capture, let alone its settings.

beaufortfrancois:
@eladalon1983 I've addressed your feedback.

beaufortfrancois:
The more I think about it, I wonder which purpose is served for this specific setting if web developers can already get the value from `getConstraints()`. Maybe we should simply remove this line from the current spec:

> As a setting, this value indicates whether or not the user agent is applying local audio playback suppression to the source.

beaufortfrancois:
Following [comment](https://github.com/w3c/mediacapture-main/issues/915#issuecomment-1303743506), we may want to actually keep `suppressLocalAudioPlayback` in `getSettings()` though.
@jan-ivar Can you review this PR?

index.html
```diff
@@ -701,8 +701,8 @@ <h2 id="constrainable-properties">
               <td><dfn class=export>suppressLocalAudioPlayback</dfn></td>
               <td>{{ConstrainBoolean}}</td>
               <td>
-                <p>As a setting, this value indicates whether or not the user
-                agent is applying <a>local audio playback suppression</a> to
+                <p>As a setting, this value indicates whether or not the constraint
+                for <a>local audio playback suppression</a> is applied to
```
jan-ivar: I'd prefer leaving the existing language alone here. As I recall, it was carefully chosen. cc @henbos 

[Local audio playback suppression](https://w3c.github.io/mediacapture-screen-share/#dfn-local-audio-playback-suppression) is already defined as a state in which _"the user agent SHOULD stop relaying audio to the local speakers"_ _(but importantly MAY NOT)._

This means we can be in this state, and we can reveal this state to JS, without revealing whether suppression is _actually_ happening.

This simultaneously satisfies the constrainable pattern, by having this state be the constrainable on/off property, while at the same allowing user agents the option to not comply in a manner that remains unobservable.

index.html
```diff
@@ -1425,6 +1426,15 @@ <h2>
                 constraint is applied (<code>true</code>) or not (<code>false</code>).
               </p>
             </dd>
+            <dt>
+               <dfn>suppressLocalAudioPlayback</dfn> of type {{boolean}}
+            </dt>
+            <dd>
+              <p>
+                Indicates whether the <a href="#dfn-suppresslocalaudioplayback">suppressLocalAudioPlayback</a>
+                constraint is applied (<code>true</code>) or not (<code>false</code>).
```
jan-ivar: I think we should just use the same language as [here](https://w3c.github.io/mediacapture-screen-share/#dfn-suppresslocalaudioplayback):
```suggestion
                Indicates whether or not the user agent is applying
                [=local audio playback suppression=] to the source.
```
The non-compliance option happens at a lower level that isn't observable. Readers can easily click on [local audio playback suppression](https://w3c.github.io/mediacapture-screen-share/#dfn-local-audio-playback-suppression) to reveal that part.

jan-ivar (changes requested):
Thanks for writing this PR to add a setting that clearly should be there judging by existing prose that already mentions it being a setting, and is therefore editorial.

beaufortfrancois:
Thank you @jan-ivar! I've addressed your feedback.

jan-ivar (approved):

index.html
eladalon1983: I think @jan-ivar meant to @mention me here, not Henrik. I was involved with suppressLocalAudioPlayback, and @henbos was involved with the adjacent control [restrictOwnAudio](https://www.w3.org/TR/screen-capture/#dfn-restrictownaudio). (A most excellent control which I am hoping we'll be able to implement soon-ish, btw).

My concern with the PR as it stands is that it produces an apparent contradiction between these two sections:
1. suppressLocalAudioPlayback of type boolean [...] Indicates whether or not the user agent is applying local audio playback suppression to the source.
2. When local audio playback suppression is applied [...] suppression MUST NOT be observable to the captured document.

I'd prefer a formulation in no1 that is more immediately clear to readers who only check one part of the document. François, please consider the following:
> suppressLocalAudioPlayback of type boolean [...] Indicates whether or not the application instructed the user agent to apply local audio playback suppression to the source.

Or even:
> suppressLocalAudioPlayback of type boolean [...] Indicates whether or not the application instructed the user agent to apply local audio playback suppression to the source. This setting does not reveal to the application whether the user agent is actually applying local audio playback suppression, as this can be overridden by the user.

beaufortfrancois: I've applied your suggestion @eladalon1983.
@jan-ivar Does that still look good to you? If so, feel free to merge.

eladalon1983 (approved):

index.html
jan-ivar: Thanks, LGTM. I suggest we merge this and let editors to work out further details in followup PRs.

FWIW I did mean to mention Henrik since this language was [modeled on](https://github.com/w3c/mediacapture-screen-share/pull/164#discussion_r622202758) restrictOwnAudio and some consistency between the two seems appropriate.

> My concern with the PR as it stands is that it produces an apparent contradiction between these two sections:
> 
> 1. suppressLocalAudioPlayback of type boolean [...] Indicates whether or not the user agent is applying local audio playback suppression to the source.
> 2. When local audio playback suppression is applied [...] suppression MUST NOT be observable to the captured document.

I don't see any contradiction here. Note that 2 is talking about the _captured_ document, not the _capturing_ document.

The constrainable pattern is a device-altering model, which very much leaks cross-origin global information by default, and I don't see anything in the current document that prevents one capturer from reading the settings applied by a different concurrent capturer, even though in practice, I believe implementations are very much implementing processing per track to maintain this privacy property whenever possible.

That said, I welcome this language (even though it's perhaps more than editorial, it seems in keeping with implementation patterns), so LGTM.

It'd be nice to apply such language consistently to other properties as well, like restrictOwnAudio perhaps, though I agree special mention seems appropriate for suppressLocalAudioPlayback since it's a property whose function is inherently global.

Lastly, the prose in this PR seems to assume the the default is `false`, even though this isn't spelled out anywhere (and the Constrainable Pattern notoriously doesn't define defaults). But I see [restrictOwnAudio](https://w3c.github.io/mediacapture-screen-share/#dom-mediatracksettings-restrictownaudio) suffers the same problem so this isn't new and seems like something editors can clarify in followups.

jan-ivar (approved):

beaufortfrancois:
@eladalon1983 @jan-ivar Please merge

eladalon1983:
The current work mode is that we merge PRs during a meeting on Thursday.

jan-ivar:
Forgot this one this morning, and it's editorial, so merging.

PR #264, opened 2023-03-28
Update how FocusChangeDisabled is set to true in case of change of focus

### LINKED ISSUE #262: Update algorithms around setFocusBehavior to take into account OS level surface pickers
In macOS, the window selection prompt might naturally unfocus the capturing web page.
As currently specified, it would always set [[[FocusChangeDisabled]]] to true, neutering setFocusBehavior.

The spec should give some leeway in that area so that [[[FocusChangeDisabled]]] is set to true when page is no longer focused for actions outside the device selection or something like this.

youennf:
The change covers potential race conditions where the change of focus task is already enqueued at the time the getDisplayMedia promise is resolved. Add a note to talk about the requirement behind these steps and that it should apply to any type of picker, whether user agent level or OS level.

Fixes https://github.com/w3c/mediacapture-screen-share/issues/262

***
<a href="https://pr-preview.s3.amazonaws.com/youennf/mediacapture-screen-share/pull/264.html" title="Last updated on Jul 6, 2023, 6:04 AM UTC (286a32d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/264/30d0ea6...youennf:286a32d.html" title="Last updated on Jul 6, 2023, 6:04 AM UTC (286a32d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -424,13 +424,12 @@
                     </p>
                   </li>
                   <li>
-                    <p>Let <var>topContext</var> be the <a>current settings object</a>'s's [=relevant global object=]'s
-                      [=associated Document=]'s [=top-level browsing context=].</p>
+                    <p>Let <var>toplevelTraversable</var> be the <a>current settings object</a>'s's [=relevant global object=]'s
+                      [=navigable=]'s [=top-level traversable=].</p>
                   </li>
                   <li>
-                    <p>Start listening to <var>topContext</var>'s change of <a data-cite="!HTML/#gains-focus">focus</a>.
-                      Whenever <var>topContext</var> is losing <a data-cite="!HTML/#gains-focus">focus</a>, stop listening
-                      to <var>topContext</var>'s change of <a data-cite="!HTML/#gains-focus">focus</a> and
+                    <p>Listen to <var>toplevelTraversable</var>'s change of <a data-cite="!HTML/#system-focus">system focus</a>.</p>
+                    <p>The first time <var>toplevelTraversable</var> is losing <a data-cite="!HTML/#system-focus">focus</a>,
                       queue a global task on the <a data-cite="!HTML/#user-interaction-task-source">user interaction task source</a>
                       given <a>current settings object</a>'s [=relevant global object=] to run the following step:
                       <ol>
--- a/index.html
+++ b/index.html
@@ -424,6 +424,29 @@
                     </p>
                   </li>
                   <li>
+                    <p>Let <var>topContext</var> be the <a>current settings object</a>'s's [=relevant global object=]'s
+                      [=associated Document=]'s [=top-level browsing context=].</p>
+                  </li>
+                  <li>
+                    <p>Start listening to <var>topContext</var>'s change of <a data-cite="!HTML/#gains-focus">focus</a>.
+                      Whenever <var>topContext</var> is losing <a data-cite="!HTML/#gains-focus">focus</a>, stop listening
+                      to <var>topContext</var>'s change of <a data-cite="!HTML/#gains-focus">focus</a> and
+                      queue a global task on the <a data-cite="!HTML/#user-interaction-task-source">user interaction task source</a>
+                      given <a>current settings object</a>'s [=relevant global object=] to run the following step:
+                      <ol>
+                        <li><p>Set <var>controller</var>.{{CaptureController/[[FocusChangeDisabled]]}} to <code>true</code>.</p></li>
+                      </ol>
+                    </p>
+                    <p class="note">
+                      These steps ensure {{CaptureController}} will not override explicit focus actions made by the user,
+                      typically if a user decides to switch to another surface shortly after starting capture.
+                      This algorithm describes what to do for surface pickers implemented by the <a>user agent</a> but the same
+                      requirement applies to surface pickers implemented outside of the <a>user agent</a>,
+                      where the loss of capturing document focus is not necessarily the signal triggering setting
+                      {{CaptureController/[[FocusChangeDisabled]]}} to <code>true</code>.
+                    </p>
+                  </li>
+                  <li>
                     <p><em>Permission Failure</em>: [=Reject=]
                     <var>p</var> with a new {{DOMException}}
                     object whose {{DOMException/name}} attribute has the
--- a/index.html
+++ b/index.html
@@ -397,6 +397,28 @@
                   <li>
                     If <var>controller</var> is not <code>null</code>, run the following steps:
                     <ol>
+                    <li>
+                      <p>Let <var>toplevelTraversable</var> be the <a>current settings object</a>'s's [=relevant global object=]'s
+                        [=navigable=]'s [=top-level traversable=].</p>
+                    </li>
+                    <li>
+                      <p>Listen to <var>toplevelTraversable</var>'s change of <a data-cite="!HTML/#system-focus">system focus</a>.</p>
+                      <p>The first time <var>toplevelTraversable</var> is losing <a data-cite="!HTML/#system-focus">focus</a>,
+                        queue a global task on the <a data-cite="!HTML/#user-interaction-task-source">user interaction task source</a>
+                        given <a>current settings object</a>'s [=relevant global object=] to run the following step:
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -423,6 +423,24 @@ <h2>
                       abort these steps. 
                     </p>
                   </li>
+                  <li>
+                    <p>If the <a>current settings object</a>'s's [=relevant global object=]'s [=associated Document=]'s
```
eladalon1983: What if focus is lost after this step executes?

youennf: The idea is that we start listening to focus loss for that controller at this exact point in time. And enqueue a task when that happens.
The `If` is indeed misleading. How about `Whenever `, or `The first time `.
Or maybe rephrase to `Listen to ... losing focus. The first time it happens, queue...`

index.html
```diff
@@ -423,6 +423,24 @@ <h2>
...
+                        <li><p>Set <var>controller</var>.{{CaptureController/[[FocusChangeDisabled]]}} to <code>true</code>.</p></li>
+                      </ol>
+                    </p>
+                    <p class="note">
+                      These steps ensure {{CaptureController}} will not override explicit focus actions made by the user,
+                      typically if a user decides to switch to another surface shortly after starting capture.
+                      This algorithm describes what to do for surface pickers implemented by the <a>user agent</a> but the same
+                      requirement applies to surface pickers implemented outside of the <a>user agent</a>,
+                      where the loss of capturing document focus is not necessarily the signal triggering setting
+                      {{CaptureController/[[FocusChangeDisabled]]}} to<code>true</code>.
```
eladalon1983: nit: Missing space after "to".

index.html
eladalon1983: The `"listen to..." ` or `"start listening to..."` variants seem promising.
And should this perhaps be done immediately after the [Prompt the user to choose](https://www.w3.org/TR/permissions/#dfn-prompt-the-user-to-choose) step?

index.html
```diff
@@ -423,6 +423,24 @@ <h2>
                       abort these steps. 
                     </p>
                   </li>
+                  <li>
+                    <p>If the <a>current settings object</a>'s's [=relevant global object=]'s [=associated Document=]'s
+                      [=top-level browsing context=] is losing <a data-cite="!HTML/#gains-focus">focus</a>,
+                      queue a global task on the <a data-cite="!HTML/#user-interaction-task-source">user interaction task source</a>
```
eladalon1983: Which source would that be? Note that focus could also be lost by a new application launching and grabbing it. There need not necessarily be a user action that triggered this focus change. (Or am I misreading this paragraph? It's a very difficult paragraph to read. Could you simplify it? Maybe a "let x be..." and then use it?)

index.html
```diff
@@ -423,6 +423,29 @@ <h2>
                       abort these steps. 
                     </p>
                   </li>
+                  <li>
+                    <p>Let <var>topContext</var> be the <a>current settings object</a>'s's [=relevant global object=]'s
+                      [=associated Document=]'s [=top-level browsing context=].</p>
```
jan-ivar: HTML demoted browsing context (still there but meant for highly specifc behavior), in favor of new language. E,g, see mediacapture-main:

If mediaDevices's [relevant global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global)'s [navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#navigable)'s [top-level traversable](https://html.spec.whatwg.org/multipage/document-sequences.html#top-level-traversable) has [system focus](https://html.spec.whatwg.org/multipage/#tlbc-system-focus), [do X]

index.html
youennf: Done

index.html
youennf: Updated to `listen to`.
As of moving the step, I think this is good where it is.
Otherwise, we would potentially set FocusChangeDisabled to true before the user chose the surface (say getDisplayMedia is called, picker is shown, and user decides to unfocus/refocus the page before answering the prompt).

index.html
youennf: The task source is the one used to queue mouse/keyboard events. Maybe we should use another task source, but I do not think it matters much in practice.

index.html
youennf: Switched to traversable and system focus.

index.html
```diff
@@ -423,6 +423,28 @@ <h2>
                       abort these steps.
```
eladalon1983: The "resolve and abort these steps" bullet right before the new bullets, seems like it ensure they'd never be reached.

index.html
```diff
@@ -423,6 +423,28 @@ <h2>
                       abort these steps. 
                     </p>
                   </li>
+                  <li>
```
jan-ivar: We never get here because the previous step says to (Resolve p and) "abort these steps", so this doesn't work.

The subsequent statements also access realm-exposed objects in parallel which is a [no-no](https://html.spec.whatwg.org/#event-loop-for-spec-authors).

Since (my understanding is) "Resolve p" from in-parallel steps is really short for: queue a task that resolves p, perhaps we meant to queue a task for the remaining steps? That would put them on main thread.

But since this seems time critical, it might be better to extract these variables ahead of the in parallel steps.

index.html
```diff
@@ -423,6 +423,28 @@ <h2>
...
+                    <p>Let <var>toplevelTraversable</var> be the <a>current settings object</a>'s's [=relevant global object=]'s
+                      [=navigable=]'s [=top-level traversable=].</p>
```
jan-ivar: This is mixing [current](https://html.spec.whatwg.org/#current) and [relevant](https://html.spec.whatwg.org/#relevant) concepts. Whenever we have a DOM object we should use what's relevant to that.

We also need to move these ahead of the in-parallel steps. Something like:

- Let mediaDevices be [=[this](https://webidl.spec.whatwg.org/#this)=].
- Let _toplevelTraversable_ be the _mediaDevice_'s [=relevant global object=]'s
                      [=navigable=]'s [=top-level traversable=].</p>

(The _mediaDevices_ temporary simplifies referencing it in prose below)

index.html
```diff
@@ -423,6 +423,28 @@ <h2>
...
+                  </li>
+                  <li>
+                    <p>Listen to <var>toplevelTraversable</var>'s change of <a data-cite="!HTML/#system-focus">system focus</a>.</p>
+                    <p>The first time <var>toplevelTraversable</var> is losing <a data-cite="!HTML/#system-focus">focus</a>,
+                      queue a global task on the <a data-cite="!HTML/#user-interaction-task-source">user interaction task source</a>
```
jan-ivar: Mixing async instructions into sequential steps can be tricky to understand. At what point does this listening begin? i.e. what is it synced to? For how long should it listen? If I call getDisplayMedia() twice, are there two listeners? etc. There's also no such concept as being able to "listen" to changes like this.

In the end the user agent is listening to itself here, so it's not clear to me what the requirements actually are. Focus changes seem highly UA specific, so maybe it's better to just spell out the intent?

Rather than constrain the user agent, it sounds like what we're trying to describe here is actually a constraint on the feature. Might this be described as an after-thought to the previous step using something like: 

- Resolve p with stream.

   If a subsequent user action causes the _toplevelTraversable_ to lose [=system focus=] before a
   [focus decision has been finalized](https://w3c.github.io/mediacapture-screen-share/#dfn-finalize-focus-decision-algorithm), the User Agent MUST set controller</var>.[[FocusChangeDisabled]] to true.
- Abort the remaining steps.

Also, I don't think the User Agent should queue a task here, since the task to finalize the focus decision may already be in the queue, and if the user already switched focus, then we want the user to always win. AFAICT the [[FocusChangeDisabled]] internal slot isn't otherwise observable to JS, so this seems fine.

I also see we already have this (is it redundant now?):

_"The user agent MAY set [[[FocusChangeDisabled]]](https://w3c.github.io/mediacapture-screen-share/#dfn-focuschangedisabled) to true at any moment based on its own logic."_

jan-ivar (changes requested):

PR #272, opened 2023-07-05
Add [=this=] to mean controller and clean up in parallel controller.[[FocusBehavior]] access.

jan-ivar:
***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/mediacapture-screen-share/pull/272.html" title="Last updated on Jul 6, 2023, 2:47 PM UTC (4a6226c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/272/30d0ea6...jan-ivar:4a6226c.html" title="Last updated on Jul 6, 2023, 2:47 PM UTC (4a6226c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -962,34 +962,34 @@
                 </li>
                 <li>
                   <p>
-                    If {{CaptureController/[[Source]]}} is <code>null</code>,
-                    set {{CaptureController/[[FocusBehavior]]}} to <var>focusBehavior</var>
+                    If [=this=].{{CaptureController/[[Source]]}} is <code>null</code>,
+                    set [=this=].{{CaptureController/[[FocusBehavior]]}} to <var>focusBehavior</var>
                     and abort these steps.
                   </p>
                 </li>
                 <li>
                   <p>
-                    If {{CaptureController/[[Source]]}} has been
+                    If [=this=].{{CaptureController/[[Source]]}} has been
                     <a data-cite="GETUSERMEDIA#source-stopped">stopped</a>,
                     [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
                   </p>
                 </li>
                 <li>
                   <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1041,19 +1041,24 @@ <h2><dfn>CaptureController</dfn></h2>
             </li>
             <li>
               <p>
-                Run the following step  [=in parallel=]:
+                Let <var>focusBehavior</var> be
+                <var>controller</var>.{{CaptureController/[[FocusBehavior]]}}.
```
eladalon1983, line 1045: 1. This seems fine. Another option would be to specify that an attempt to change `[[FocusBehavior]]` when `[[IsBound]]` is `true` should result in an error. The benefit of that other approach would be that it would prevent incorrect usage of CaptureController by Web developers. Wdyt?
2. Should we also cache `[[Source]]` in case dynamic switching changes it in future changes? An in that vein - shall we follow up in a separate PR and specify that changes derived from [dynamic switching](https://www.w3.org/TR/screen-capture/#dom-displaymediastreamoptions-surfaceswitching) change the `[[Source]]`? @youennf, note this potentially affects Safari-on-macOS.

eladalon1983: I've taken a bit of a closer look, and it's not clear that there is a race, because we break out of both [finalize focus decision algorithm](https://www.w3.org/TR/screen-capture/#dfn-finalize-focus-decision-algorithm) and [setFocusBehavior()](https://www.w3.org/TR/screen-capture/#dom-capturecontroller-setfocusbehavior) before reaching the part where the race was suspected. But I'm OK with merging this PR for clarification's sake.

(I'd also like to continue with the discussion over #2, although possibly I should file a separate issue for that.)

eladalon1983 (approved):

PR #274, opened 2023-08-25
Define monitorTypeSurfaces

### LINKED ISSUE #261: Allow apps to avoid riskier display-surface types
Some video-conferencing tools offer admin settings that allow tuning what the user may do. For example, some admins allow sharing of screens when all participants are employees, but restrict to window-sharing in calls that have external participants (under the assumption that accidental leaks would be more costly in such scenarios).

Native apps allow such fine-grained control. Web apps... almost. Web apps can call `getDisplayMedia()` and pray. If the user chooses to share a monitor, these apps can stop the capture and chide the user for making an improper choice - a bad experience for all involved (remote participants included, as they sit and listen to the user complain about the Web app).

It would be better if we allowed Web apps to remove the `monitor` option from what is offered to the user. We could go with an API shape that's similar to the existing [selfBrowserSurface](https://w3c.github.io/mediacapture-screen-share/#dom-displaymediastreamoptions-selfbrowsersurface) option.

```webidl
enum MonitorTypeSurfacesEnum {
  "include",
  "exclude"
};

dictionary DisplayMediaStreamOptions {
  MonitorTypeSurfacesEnum monitorTypeSurfaces;
};
```

Note that I propose a shape that only allows removing monitors, which is the riskiest option. Whether windows should be exclude-able is its own discussion. I think that if we ever decide that these too should be exclude-able, we could specify that one cannot exclude windows without also excluding monitors, lest the user be nudged towards monitors.

If we also specify [getSupportedOptions()](https://github.com/w3c/mediacapture-screen-share/issues/260), apps would be able to know ahead of time whether such shaping is possible, btw.

CC @Coread, who is interested in such a mechanism. (He might have different opinions about the API shape and some finer points, though. Please let us know.)

---
It should be mentioned that discussions in this vein arose somewhat often. I believe this particular variant has not been rejected before. The latest discussion on an adjacent topic was #209, which culminated with the aforementioned [selfBrowserSurface](https://w3c.github.io/mediacapture-screen-share/#dom-displaymediastreamoptions-selfbrowsersurface) being specified - a success-story which I hope to repeat.

eladalon1983:
Fixes #261.

***
<a href="https://pr-preview.s3.amazonaws.com/eladalon1983/mediacapture-screen-share/pull/274.html" title="Last updated on Aug 31, 2023, 2:44 PM UTC (525c995)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/274/1f8eb00...eladalon1983:525c995.html" title="Last updated on Aug 31, 2023, 2:44 PM UTC (525c995)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1287,8 +1287,8 @@ dictionary DisplayMediaStreamOptions {
                 <dd>
                   <p>
                     If present, signals whether the application would like the user agent
-                    to offer the user the option to choose [=display surfaces=] whose type
-                    is [=display surface/monitor=].
+                    to include [=display surfaces=] whose type is [=display surface/monitor=]
+                    among the choices offered to the user. The user agent MAY ignore this hint.
                   </p>
                   <div class="note">
                     <p>
--- a/index.html
+++ b/index.html
@@ -1164,9 +1164,9 @@
         <section>
           <h2><dfn>MonitorTypeSurfacesEnum</dfn></h2>
           <p>
-            Describes whether an application invoking {{MediaDevices/getDisplayMedia()}}
-            would like the user agent to offer the user an option to dynamically switch
-            the source [=display surface=] during the capture.
+            If present, described whether the application would like the user agent
+            to offer the user the option to choose [=display surfaces=] whose type
+            is [=display surface/monitor=].
           </p>
           <pre class="idl">
             enum MonitorTypeSurfacesEnum {
--- a/index.html
+++ b/index.html
@@ -1292,10 +1292,9 @@ dictionary DisplayMediaStreamOptions {
                   </p>
                   <div class="note">
                     <p>
-                      {{monitorTypeSurfaces}} is just a hint. The user agent may still offer to the user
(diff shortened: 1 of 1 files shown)


eladalon1983:
CC @Coread

index.html
```diff
@@ -1240,6 +1281,24 @@ <h2>Dictionary <a class="idlType">DisplayMediaStreamOptions</a>
...
+                  <p>
+                    If present, signals whether the application would like the user agent
+                    to offer the user the option to choose [=display surfaces=] whose type
+                    is [=display surface/monitor=].
+                  </p>
+                  <div class="note">
+                    <p>
+                      {{monitorTypeSurfaces}} is just a hint. The user agent may still offer to the user
+                      the option to capture [=display surfaces=] of type [=display surface/monitor=].
+                      Applications are therefore encouraged to still check the {{displaySurface}} setting
```
```suggestion
                      Applications are therefore encouraged to still check the {{MediaTrackSettings/displaySurface}} setting
```

jan-ivar (approved):

PR #283, opened 2023-11-08
Define the windowAudio option

### LINKED ISSUE #7: Avoid user-confusion by not offering undesired audio sources
User agents may offer audio to be captured alongside video, if the application specifies `{audio: true}`, or maps `audio` to anything else that's different from `false`. But **not all audio is created alike**. Consider video-conferencing applications. Tab-audio is often useful, and can be shared with remote participants. But system-audio includes participants' own audio, and it is NOT desirable to share it back.

State of the art? VC applications can ask for "audio", use it if it's tab-audio, and discard it otherwise. This works, but it's sub-optimal. The user is left confused. The user wanted to share system-audio, the user was offered to share-system, the user **explicitly approved** sharing system-audio - and now remote participants are telling the user that they can't, in fact, hear the system-audio. Now how messed up is that?!

It'd be better if we allowed the application to **ask for less**. Allow the application to ask for audio only in conjunction with surfaces of type X. (For example, with constraints.)

I think we should debate this in order:
1. Desirability. Are we in agreement that it is desirable to allow the application to ask for less?
2. Mechanism. Provided that we agree on no1, how do we want to shape the control surface?

### LINKED ISSUE #219: Avoid user-confusion by not offering undesired audio sources
User agents may offer audio to be captured alongside video, if the application specifies `{audio: true}`, or maps `audio` to anything else that's different from `false`. But **not all audio is created alike**. Consider video-conferencing applications. Tab-audio is often useful, and can be shared with remote participants. But system-audio includes participants' own audio, and it is NOT desirable to share it back.

State of the art? VC applications can ask for "audio", use it if it's tab-audio, and discard it otherwise. This works, but it's sub-optimal. The user is left confused. The user wanted to share system-audio, the user was offered to share-system, the user **explicitly approved** sharing system-audio - and now remote participants are telling the user that they can't, in fact, hear the system-audio. Now how messed up is that?!

It'd be better if we allowed the application to **ask for less**. Allow the application to ask for audio only in conjunction with surfaces of type X. (For example, with constraints.)

I think we should debate this in order:
1. Desirability. Are we in agreement that it is desirable to allow the application to ask for less?
2. Mechanism. Provided that we agree on no1, how do we want to shape the control surface?

eladalon1983:
Fixes w3c/mediacapture-screen-share-extensions#7.

CC @o1ka, @jan-ivar, @youennf

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/pull/283.html" title="Last updated on Feb 23, 2024, 8:51 AM UTC (4b7a95f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/283/661c82b...4b7a95f.html" title="Last updated on Feb 23, 2024, 8:51 AM UTC (4b7a95f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -301,6 +301,16 @@
 
 
             <li>
+              <p>If <var>options</var>.{{DisplayMediaStreamOptions/windowAudio}}
+              is set to {{WindowAudioPreferenceEnum/"exclude"}}, and
+              <var>options</var>.{{DisplayMediaStreamOptions/systemAudio}}
+              is either unset or set to {{SystemAudioPreferenceEnum/"include"}},
+              return a promise [=reject|rejected=] with a newly [=exception/created=]
+              {{TypeError}}.</p>
+            </li>
+
+
+            <li>
               <p>For each [= map/exist | existing =] member in
               <var>constraints</var> whose value, <var>CS</var>, is a
               dictionary, run the following steps:</p>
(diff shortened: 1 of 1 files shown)


jan-ivar:
Sounds reasonable to me.

kvrban:
I run a platform for live streaming of games. (similar to twitch) and use webrtc / getDisplayMedia for this.
So far the limitation is that the user has the possibility to select a separate window which he wants to share, but then without sound. Or he shares the entire screen content + the entire system sound.

Do I understand correctly that with the windowAudio option it will be possible to share a separate window + the sound from only this window? That would be ideal and exactly what a user expects.

eladalon1983:
> Do I understand correctly that with the windowAudio option it will be possible to share a separate window + the sound from only this window?

Sorry, you appear to have misunderstood this PR. Here we deal with the spec. It is already possible for users to share a window+audio - per the spec. That various browsers don't offer that capability in practice is a separate issue, to be discussed on those browsers' specific bug trackers.

youennf:
Closing since it is now obsolete.

eladalon1983:
For posterity, this issue was handled by w3c/mediacapture-screen-share/pull/323.

PR #289, opened 2023-11-30
Added DisplaySurfaceChangeCallback

### LINKED ISSUE #4: Auto-pause capture when user switches captured content
Both Chrome and Safari now allow the user to change what surface is captured.

<p align="center">
   I was too lazy for alt-text.
I was too lazy for alt-text.
</p>

That's obviously great stuff. Can we make it better still?
* What if an application intends to do some processing that depends on the captured content?
* What if an application wants to set different constraints, for instance when capturing a window vs. when capturing a screen?
* What if the application intends to save different surfaces to different files, and wants to start appending to a new file whenever the user changes the source?

So I propose that we add **two things**:
1. A control which allows an application to instruct the browser - whenever the user changes what surface is shared, pause the track (by setting [enabled](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/enabled) to false).
2. Fire an event whenever this happens. (With the general expectation that the application will apply new processing and then set `enabled` back to true.)

Possibly the two can be combined, by specifying that setting an event handler signals that the pausing behavior is desired (@alvestrand's idea).

Another natural extension of this idea is to also apply it when a captured tab undergoes cross-origin navigation of the top-level document. When that happens, some applications might wish to stop capture momentarily and (in-content) prompt the user - "do you still want to keep sharing?"

Relevant previous discussion [here](https://github.com/w3c/mediacapture-screen-share-extensions/issues/9).

### LINKED ISSUE #255: Auto-pause capture when user switches captured content
Both Chrome and Safari now allow the user to change what surface is captured.

<p align="center">
   I was too lazy for alt-text.
I was too lazy for alt-text.
</p>

That's obviously great stuff. Can we make it better still?
* What if an application intends to do some processing that depends on the captured content?
* What if an application wants to set different constraints, for instance when capturing a window vs. when capturing a screen?
* What if the application intends to save different surfaces to different files, and wants to start appending to a new file whenever the user changes the source?

So I propose that we add **two things**:
1. A control which allows an application to instruct the browser - whenever the user changes what surface is shared, pause the track (by setting [enabled](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/enabled) to false).
2. Fire an event whenever this happens. (With the general expectation that the application will apply new processing and then set `enabled` back to true.)

Possibly the two can be combined, by specifying that setting an event handler signals that the pausing behavior is desired (@alvestrand's idea).

Another natural extension of this idea is to also apply it when a captured tab undergoes cross-origin navigation of the top-level document. When that happens, some applications might wish to stop capture momentarily and (in-content) prompt the user - "do you still want to keep sharing?"

Relevant previous discussion [here](https://github.com/w3c/mediacapture-screen-share/issues/182#issuecomment-855976799).

tovepet:
Fixes w3c/mediacapture-screen-share-extensions#4.

***
<a href="https://pr-preview.s3.amazonaws.com/tovepet/mediacapture-screen-share/pull/289.html" title="Last updated on Oct 2, 2024, 1:03 PM UTC (7c6434f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/289/d864164...tovepet:7c6434f.html" title="Last updated on Oct 2, 2024, 1:03 PM UTC (7c6434f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -409,25 +409,37 @@
 
                 <li>
                   <p>[=Prompt the user to choose=] a display device, for a
-                  {{PermissionDescriptor}} with its
-                  {{PermissionDescriptor/name}} set to "display-capture",
-                  resulting in a set of provided media.</p>
+                    {{PermissionDescriptor}} with its
+                    {{PermissionDescriptor/name}} set to "display-capture",
+                    resulting in a <dfn>capture-source-set</dfn>,
+                    <var>sources</var>, that meets the following
+                    requirements:</p>
 
 
-                  <p>The provided media MUST include precisely one video
-                  track.</p>
+                  <ol>
+                    <li>
+                      <p>The [=capture-source-set=] MUST include precisely one
+                      video track.</p>
+                    </li>
 
 
-                  <p>The provided media MUST include at most one audio track.
-                  This audio track MUST NOT be included if audio was not
-                  specified in <var>requestedMediaTypes</var>, or if it was
-                  specified as <code>false</code>.</p>
+                    <li>
+                      <p>The [=capture-source-set=] MUST include at most one
+                        audio track. This audio track MUST NOT be included if
+                        audio was not specified in
+                        <var>requestedMediaTypes</var>, or if it was specified
+                        as <code>false</code>.</p>
+                    </li>
 
 
-                  <p>The devices chosen MUST be the ones determined by the
-                  user. Once selected, the source of a {{MediaStreamTrack}}
-                  MUST NOT change, unless the user permits it through their
-                  interaction with the user agent.</p>
+                    <li>
+                      <p>The sources in the [=capture-source-set=] MUST
+                        represent the devices determined by the user. Once
+                        selected, the source of a {{MediaStreamTrack}} MUST NOT
+                        change, unless the user permits it through their
+                        interaction with the user agent.</p>
+                    </li>
+                  </ol>
 
 
                   <p>User agents are encouraged to warn users against sharing
@@ -439,9 +451,9 @@
 
 
                   <p>If the result of the request is
-                  {{PermissionState/"granted"}}, then for each device that is
-                  sourcing the provided media, using a stable and private id
-                  for the device, <var>deviceId</var>, set
+                  {{PermissionState/"granted"}}, then for each device with a
+                  source in [=capture-source-set=], using a stable and private
+                  id for the device, <var>deviceId</var>, set
                   [[\devicesLiveMap]]<var>[deviceId]</var> to
                   <code>true</code>, if it isn’t already <code>true</code>, and
                   set the [[\devicesAccessibleMap]]<var>[deviceId]</var> to
@@ -512,7 +524,8 @@
                 <li>
                   <p>This invocation of {{MediaDevices/getDisplayMedia()}} is
                   now considered to have produced a new
-                  <dfn>capture-session</dfn>.</p>
+                  <dfn>capture-session</dfn>, with the associated
+                  [=capture-source-set=], <var>sources</var>.</p>
                 </li>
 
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -490,7 +502,8 @@ <h2><dfn>MediaDevices</dfn> Additions</h2>
                 <li>
                   <p>This invocation of {{MediaDevices/getDisplayMedia()}} is
                   now considered to have produced a new
-                  <dfn>capture-session</dfn>.</p>
+                  <dfn>capture-session</dfn>, with the associated
+                  [=capture-source-set=], <var>sources</var>.</p>
                 </li>
```
eladalon1983, line 529: Should the part below, which dealt with `controller.[[Source]]`, be updated to reference the video source from `capture-source-set` rather than (as it does now) the "video track's source"?
(GitHub didn't let me attach this comment in the desired location.)

index.html
```diff
@@ -646,6 +659,106 @@ <h2 id="constraints">Unconstrained Display Surface Selection</h2>
...
 
+    <section>
+      <h2 id="changing-display-surfaces">Changing Display Surfaces</h2>
+
+
+      <p>
+        The user agent MUST NOT change the [=display surface=]
+        associated with a [=capture-session=] unless the user has
+        explicitly indicated that they want this change to be
+        performed by interacting with user agent or operating system.
```
eladalon1983, line 692: nit: Could you make editorial changes here so as to discourage anyone from misreading "by interacting" as binding to the wrong part of the sentence? Maybe "...performed. Users can do this by interacting with..."

index.html
```diff
@@ -646,6 +659,106 @@ <h2 id="constraints">Unconstrained Display Surface Selection</h2>
...
+            <var>newSources</var> MUST fulfill the same requirements as the
+            [=capture-source-set=] formed when {{MediaDevices/getDisplayMedia}}
+            is called.</p>
+        </li>
+
+
+        <li>
+          <p>
+            Let <var>newSources</var> replace the [=capture-source-set=]
+            associated with the [=capture-session=]
```
eladalon1983: nit: Missing full-stop.

index.html
```diff
@@ -646,6 +659,106 @@ <h2 id="constraints">Unconstrained Display Surface Selection</h2>
...
+            [=capture-source-set=] associated with the [=capture-session=].
+          </p>
+        </li>
+
+
+         <li>
+           <p>Queue tasks to stop all tracks connected to sources in the
+             [=capture-source-set=] associated with the [=capture-session=].
+          </p>
+        </li>
```
eladalon1983: The order in the PR as it currently stands:
1. Stop delivering frames.
2. Queue tasks to end the old tracks.
3. Synchronously make changes to internal slots.
4. Synchronously invoke the callback.

Perhaps it should be:
1. Stop delivering frames.
2. Queue task to:
  a. Make changes to internal slots.
  b. Invoke the callback.
3. Queue tasks to end the old tracks.

Wdyt?

eladalon1983 (commented):
Generally LGTM, but some suggestions.

index.html
tovepet: I've grouped the synchronous steps into a queued task, and moved the ending of tracks to the end (+some other minor improvements)

index.html
```diff
@@ -668,6 +681,110 @@ <h2 id="constraints">Unconstrained Display Surface Selection</h2>
     </section>
 
 
+    <section>
+      <h2 id="changing-display-surfaces">Changing Display Surfaces</h2>
+
+
+      <p>
+        The user agent MUST NOT change the [=display surface=]
```
youennf, line 689: I don't think we can validate this, I would tend to reword so that we do not use MUST NOT.

index.html
```diff
@@ -668,6 +681,110 @@ <h2 id="constraints">Unconstrained Display Surface Selection</h2>
...
+        associated with a [=capture-session=] unless the user has
+        explicitly indicated that they want this change to be
+        performed by interacting with user agent or operating system.
+
+        If the user indicates a change of [=display surface=], the
```
youennf, line 694: We probably need to specify the order of the callback execution and the `configurationchange` event at MediaStreamTrack.
I would tend to have the event be executed after the callback.

index.html
```diff
@@ -668,6 +681,110 @@ <h2 id="constraints">Unconstrained Display Surface Selection</h2>
...
+
+        <li>
+          <p>If <var>controller.{{CaptureController/[[DisplaySurfaceChangeCallback]]}}</var>
+            is null, abort these steps.</p>
+        </li>
+
+
+        <li>
+          <p>Stop delivering frames from all sources in the
+            [=capture-source-set=] associated with the [=capture-session=].
```
youennf, line 714: If audio source is not changing, should we allow audio to flow?

index.html
```diff
@@ -668,6 +681,110 @@ <h2 id="constraints">Unconstrained Display Surface Selection</h2>
...
+        </li>
+
+
+        <li>
+          <p>Let <var>newSources</var> be the set of sources selected by the
+            user when the user chose a new [=display surface=].</p>
+
+
+          <p>
+            <var>newSources</var> MUST fulfill the same requirements as the
```
youennf, line 725: It is not really clear to me what this MUST is about.

index.html
```diff
@@ -668,6 +681,110 @@ <h2 id="constraints">Unconstrained Display Surface Selection</h2>
...
+      <ol>
+        <li>
+          <p>If the [=capture-session=] does not have an associated
+            {{CaptureController}}, <var>controller</var>, abort these steps.</p>
+        </li>
+
+
+        <li>
+          <p>If <var>controller.{{CaptureController/[[DisplaySurfaceChangeCallback]]}}</var>
+            is null, abort these steps.</p>
```
youennf, line 708: I think we should do this check later inside a task queued since CaptureController/[DisplaySurfaceChangeCallback] can be changed at any point in time by JS, so the checks should be done within the context event loop.

index.html
```diff
@@ -668,6 +681,110 @@ <h2 id="constraints">Unconstrained Display Surface Selection</h2>
...
+
+          <p>
+            <var>newSources</var> MUST fulfill the same requirements as the
+            [=capture-source-set=] formed when {{MediaDevices/getDisplayMedia}}
+            is called.</p>
+        </li>
+
+
+        <li>
+          <p>Queue a task to:</p>
```
youennf, line 732: Will it have a link to the definition?
We should probably start using a task source as well.

index.html
```diff
@@ -668,6 +681,110 @@ <h2 id="constraints">Unconstrained Display Surface Selection</h2>
...
+            <li>
+              <p>Let <var>stream</var> be a new {{MediaStream}} with a new
+                {{MediaStreamTrack}} for each source in <var>newSources</var>.
+            </li>
+
+
+            <li>
+              <p>Run the [=ApplyConstraints algorithm=] on all tracks in
+                <var>stream</var> with the appropriate constraints. Should
+                this fail, abort these steps.</p>
```
youennf, line 751: Why should we fail, user wants to switch, it is ok for the web page to update its constraints as part of the switch.

index.html
```diff
@@ -668,6 +681,110 @@ <h2 id="constraints">Unconstrained Display Surface Selection</h2>
...
+              <p>Invoke
+                <var>controller</var>.{{CaptureController/[[DisplaySurfaceChangeCallback]]}}
+                with <var>stream</var> as the argument.</p>
+            </li>
+          </ol>
+        </li>
+
+        <li>
+          <p>Queue tasks to end all tracks connected to sources in the
+            [=capture-source-set=] associated with the [=capture-session=].
```
youennf, line 781: I think the default should be to leave the tracks alive, the future option being for the web page to ask for tracks to be ended.
The principle is that a web page can easily implement ending of the tracks, but it cannot implement the reverse.

index.html
```diff
@@ -1294,6 +1449,45 @@ <h2><dfn>CaptureController</dfn>
...
+
+
+            <ol>
+              <li>
+                <p>Let <var>callback</var> be the method's first argument.</p>
+              </li>
+
+
+              <li>
+                <p>If [=this=].{{CaptureController/[[IsBound]]}} is
```
youennf, line 1469: It seems restrictive to disallow setting the callback more dynamically.
Why should we add this restriction?

index.html
```diff
@@ -1294,6 +1449,45 @@ <h2><dfn>CaptureController</dfn>
...
+                [=exception/throw=] an "{{InvalidStateError}}"
+                {{DOMException}}.</p>
+              </li>
+
+
+              <li>
+                <p>If
+                  [=this=].{{CaptureController/[[DisplaySurfaceChangeCallback]]}}
+                  is not <code>null</code>, [=exception/throw=] an
+                  "{{InvalidStateError}}" {{DOMException}}.</p>
```
youennf, line 1480: I am not sure why we should not allow to overwrite the callback either.

jan-ivar: This seems different from https://github.com/w3c/mediacapture-screen-share-extensions/issues/4#issuecomment-2373957275:
> > Something like `captureController.processSourceSwitch(stream => { ... });` or `captureController.processSourceSwitch(null);`
> 
> I’m fine with this.

PR #322, opened 2025-05-21
getDisplayMedia's promise confers transient activation when resolved

eladalon1983:
When invoked, getDisplayMedia() returns a promise. If the user chooses to share a tab, a window or a screen, that promise is resolved. We specify that when that happens, this confers renewed transient activation in the document which has called getDisplayMedia().

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/pull/322.html" title="Last updated on May 22, 2025, 2:34 PM UTC (314508a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/322/8d0f388...314508a.html" title="Last updated on May 22, 2025, 2:34 PM UTC (314508a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -541,7 +541,7 @@
                     </li>
                     <li>
                       Set the <a data-cite="html/#last-activation-timestamp">last activation timestamp</a>
-                      to the <a data-cite="hr-time#dfn-current-high-resolution-time">current high resolution time</a>.
+                      to the [=current high resolution time=].
                     </li>
                     <li>
                       <p>[=Resolve=] <var>p</var> with <var>stream</var>.</p>
--- a/index.html
+++ b/index.html
@@ -540,6 +540,14 @@
                       </ol>
                     </li>
                     <li>
+                      Set the <a data-cite="html/#last-activation-timestamp">last activation timestamp</a>
+                      to the <a data-cite="hr-time#dfn-current-high-resolution-time">current high resolution time</a>.
+                    </li>
+                    <li>
+                      Set the <a data-cite="html/#last-activation-timestamp">last activation timestamp</a>
+                      to the <a data-cite="hr-time#dfn-current-high-resolution-time">current high resolution time</a>.
+                    </li>
+                    <li>
                       <p>[=Resolve=] <var>p</var> with <var>stream</var>.</p>
                     </li>
                   </ol>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -539,6 +539,10 @@ <h2><dfn>MediaDevices</dfn> Additions</h2>
                         </li>
                       </ol>
                     </li>
+                    <li>
+                      Set the <a data-cite="html/#last-activation-timestamp">last activation timestamp</a>
```
dontcallmedom, line 543: as far as I can determine, at the moment, the only web spec that sets the last activation timestamp is the HTML spec itself as part of the [activation notification](https://html.spec.whatwg.org/multipage/interaction.html#activation-notification), and that algorithm is somewhat involved, taking into account nested navigables.

I suspect this will need further refining (unless we can simply invoke the "activation notification" algorithm as is)

youennf (approved):

jan-ivar (approved):

index.html
youennf: We will send an issue to HTML repo and will discuss with HTML folks

eladalon1983: An adjacent issue I have just filed is [here](https://github.com/whatwg/html/issues/11330), as per Youenn's and Jan-Ivar's request during the interim meeting. For the issue Dom raises wrt specifying activation, a separate issue might have to be filed.

mustaqahmed: Unfortunately changing the timestamp through a JS API is unacceptable from security/abusability perspective.  Let's continue the discussion in https://github.com/whatwg/html/issues/11330, thanks for filing the issue.

eladalon1983: @mustaqahmed, please note that the timestamp is NOT changed through a JS API. It is changed from the user's interaction with UA-level UX that was shown to the user in response to the API.

eladalon1983: That is, you could call the JS API infinitely many times and not get the timestamp updated even once.

mustaqahmed: Oh, right!  I missed that Step 3 "Prompt the user" skips the timestamp change using "goto" steps when user interaction is missing or the user declines to give permission.  I have no major objection here.  I will comment on whatwg/html#11330 shortly with a suggestion.

PR #328, opened 2026-02-06
Add algorithm for browser automation.

lutien:
Hi here!
The Browser Testing and Tools Working Group would like to introduce the API in the WebDriver BiDi protocol to record a screen video during the test or automation scripts. The current suggestion is to base these commands on the `getDisplayMedia` API to produce the video stream (see https://github.com/w3c/webdriver-bidi/issues/636). 
Since we probably require bypassing a lot of pre-checks (user activation, permission, display surface selection), in this PR I've rather extracted the required logic in the separate algorithm, but please let me know if you have a better idea how to reuse the logic.

***
<a href="https://pr-preview.s3.amazonaws.com/lutien/mediacapture-screen-share/pull/328.html" title="Last updated on Feb 6, 2026, 4:04 PM UTC (2764a27)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-screen-share/328/81ad880...lutien:2764a27.html" title="Last updated on Feb 6, 2026, 4:04 PM UTC (2764a27)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -248,6 +248,11 @@
 
 
             <li>
+              <p>Let <var>options</var> be the method's first argument.</p>
+            </li>
+
+
+            <li>
               <p>Let <var>controller</var> be
               <var>options</var>.<code>controller</code> if present, or
               <code>null</code> otherwise.</p>
@@ -282,11 +287,6 @@
               [=transient activation=], return a promise <a>rejected</a> with a
               {{DOMException}} object whose {{DOMException/name}} attribute has
               the value {{InvalidStateError}}.</p>
-            </li>
-
-
-            <li>
-              <p>Let <var>options</var> be the method's first argument.</p>
             </li>
 
 
--- a/index.html
+++ b/index.html
@@ -291,65 +291,8 @@
 
 
             <li>
-              <p>Let <var>constraints</var> be
-              <code>[</code><var>options</var>.<code>audio</code>,
-              <var>options</var>.<code>video</code><code>]</code>.</p>
-            </li>
-
-
-            <li>
-              <p>If <code>constraints.video</code> is <code>false</code>,
-              return a promise [=reject|rejected=] with a newly
-              [=exception/created=] {{TypeError}}.</p>
-            </li>
-
-
-            <li>
-              <p>For each [= map/exist | existing =] member in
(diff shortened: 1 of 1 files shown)


tovepet:
@guidou could you take a look please?

guidou:
cc @youennf @jan-ivar

index.html
```diff
@@ -2574,5 +2586,135 @@ <h2>Feedback and Interface During Capture</h2>
...
     </section>
   </section>
+
+
+  <section>
+    <h2 id="automation">Automation</h2>
+
+
+    <p>The <dfn class="export">capture a browser tab</dfn> algorithm is used to capture a
+    browser tab for automation purposes, such as by [[WEBDRIVER-BIDI]].</p>
```
alvestrand, line 2596: Should add something like:

This algorithm can only be executed once adequate permission to capture the tab has been ensured. The algorithm itself, unlike the getDisplayMedia algorithm, does not do any permission checking.

index.html
```diff
@@ -2574,5 +2586,135 @@ <h2>Feedback and Interface During Capture</h2>
...
+
+
+        <ol>
+          <li>
+            <p>For each media type <var>T</var> in
+            <var>requestedMediaTypes</var>,</p>
+
+
+            <ol>
+              <li>
```
alvestrand, line 2626: There's a lot of repeated text in here with getDisplayMedia. Can we make the algorithms common?

index.html
```diff
@@ -2574,5 +2586,135 @@ <h2>Feedback and Interface During Capture</h2>
       any active capture is advisable.</p>
     </section>
   </section>
+
+
+  <section>
+    <h2 id="automation">Automation</h2>
```
jan-ivar, line 2592: I wonder if a better place for this might be https://www.w3.org/TR/mediacapture-viewport/

That seems like a better fit since it avoids dealing with the prompt and already has language for finding the "video inferred by browsing context". We'd only need to deal with/remove its document policy restriction.

lutien: That sounds reasonable to me. I'm going to try integrating it there then.

lutien:
Closing in favor of https://github.com/w3c/mediacapture-viewport/pull/34.

